PR #130, opened 2015-01-17
Update WordPress version to current (4.1 at January 2015)

kcristiano:
Update WordPress version to current one as 3.7 is about 15 months old.

### DIFF
--- a/app/config/wp-demo/download.sh
+++ b/app/config/wp-demo/download.sh
@@ -4,7 +4,7 @@
 
 ###############################################################################
 
-[ -z "$CMS_VERSION" ] && CMS_VERSION=3.7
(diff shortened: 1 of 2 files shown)


totten:
I tried this out, and it seems like Civi works. The default theme is a bit... bland... http://imgur.com/GbUof75 Does that seem OK? Or is there some wp-cli command we can call to get something nicer?

kcristiano:
We can use wp theme activate twentythirteen (which will give us the same theme as 3.7) or wp theme activate twentyfourteen (the previous theme).  I think that using twentyfourteen is safer as it will be bundled longer.  Then again, we can always update when it gets removed.  I'd expect that it will get removed for twentysixteen which should hit Q4 of 2015 or Q1 of 2016.

PR #136, opened 2015-03-06
HHVM compatibility when using "php -r"

adamwight:
HHVM tries to interpret command-line arguments instead of passing them to
the script.  A double-dash ensures that arguments are passed to the script.
This is compatible with Zend PHP.

Change-Id: Ie6d8109f9324564d02c3754b6acf1e5721ca91c6


totten:
Thanks. Confirmed that this still works in Zend.

PR #143, opened 2015-04-09
Give instructions on how to use civicrm-buildkit via composer

generalredneck:
Updated Composer.json with a few tweaks. Typically, you want to update dependencies unless there is a known reason NOT to, and then you pin them. I found it difficult to do so with these packages because most have been forked.

Consider adding civicrm/civicrm-buildkit to packagist.org as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ would need to re-run ***export***.
 If you want to set up buildkit using composer, do the following...
 
 * Add the following to ~/.composer/composer.json
+  
   ```
   {
     "minimum-stability": "dev",
@@ -94,6 +95,7 @@ If you want to set up buildkit using composer, do the following...
     }
   }
   ```
+  
 * Run `composer global update`
 * Add `~/.composer/vendor/bin` to your PATH varariable like so:
   `export PATH=$HOME/.composer/vendor/bin:$PATH`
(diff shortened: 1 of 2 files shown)


totten:
What happens to the data folders (eg "build/" and "app/private") when you use this approach?

totten:
This PR was a neat idea, but the goal of installing all commands via `composer require` or `composer global require` will probably never work. :(

The main problem is that PHP can only load one version of a given library at runtime, and `composer` is therefore designed to download one version of each library. (This is actually a sensible behavior when focusing on a single application... like most PHP developers do...) However, if you have two PHP CLI apps (e.g. `drush` and `wp-cli`) which use the same library (e.g. Symfony Console) but with different major versions (Symfony 2.x vs 3.x), then `composer` cannot resolve the conflict. (In theory, you could rework composer's linker to mitigate some edge-cases, but that's a big change - and still faces some hard limits from the PHP runtime.) 

For the foreseeable future, the most practical resolution is to download PHP CLI tools as PHAR executables (each CLI tool is statically-linked to its libraries).

A secondary problem is that buildkit includes some tools aren't based on PHP or distributed via packagist -- e.g.  `karma` (Javascript) and `hub` (Go-based; native binaries).

PR #151, opened 2015-05-26
Remove warning messages from hr15 + add civicrm_entity

gregmeszaros:
Remove warning messages from hr15 + add civicrm_entity module, set default theme for drupal
- warning messages fixed too


totten:
Reads OK; deferring to your judgment.

PR #153, opened 2015-06-20
civibuild - Allow configuration overrides via app/civibuild.conf

totten:


eileenmcnaughton:
Somewhat related is this PR that allows overriding of AMP dir with an environment variable - https://github.com/totten/amp/pull/29/files but I'm not sure about the fact it's a different approach. The system variable seems slightly simpler in that it can be before buildkit installation starts - as part of the server setup script - but I don't suppose it's that big a difference

totten:
I think they're mostly orthogonal.

One *could* use them together, but that's not necessary. Let's say you want to customize `AMPDIR`. Somewhere, you need to place a statement like `export AMPDIR=/var/lib/amp`. There are several choices:

 * `$HOME/.bashrc` -- Customize AMPDIR for some particular user(s) (but necessarily all users).
 * `/etc/profile.d/amp.sh` -- Customize AMPDIR for all users.
 * `/srv/buildkit/app/civibuild.conf` -- Customize AMPDIR for all users, but only in the context of `civibuild`. If someone calls `amp` as a standalone command, don't customize it.

Regardless of what you do with `AMPDIR`, there are other variables that might be nice to customize in `civibuild.conf`, such as `BLDDIR` or `ADMIN_PASS`.

eileenmcnaughton:
I won't say I have everything working as I want just yet - but this PR seems to me to be an improvement - I am +1 on merging it

PR #155, opened 2015-06-30
add distmaker notes

eileenmcnaughton:
So, I got on quite well towards creating a fuzion version of buildkit. Some limitations I'd like to look at 

1) the links in the latest dir include the date. Since I'd like a consistent url I would want the links to specific tarballs in the LATEST dir to be (e.g) 

civicrm-4.6.5-drupal.tar.gz

rather than

civicrm-4.6.5-drupal-20150630.tar.gz

The tarballs themselves would not change but it would mean there would be a url that always retrieved the latest

2) point versions / release candidates. 

My preference is to have  a branch for each point release (our convention has been 4.4.15rc1 etc). The main reason I like this is that each point release (or so) I rebase against upstream & the new point release generally has less, cleaner patches. As part of this process I make sure we aren't accumulating too many patches that have not been managed into the upstream repo, so it's partly about being a good citizen & ensuring patches get into core & partly about protecting our customers from regressions when they upgrade. Of course we do try to upstream patches whenever we do them - but some take longer than others to get into core for various reasons.

However, to do it with this system we would need to ditch the hybrid I settled on of having a branch for the repos we alter (core & drupal) & also create a branch per repo for joomla, & wordpress. (packages sits somewhere in the middle since we rarely patch it). 

From our point of view options are
1) just stick with one 4.6 branch, don't rebase & manage merge conflicts  as they happen & lose our regular upstreaming checks.

2) stick with one 4.6 patch & break Torrance's heart by doing push -f to it when we rebase & upstream

3) create dummy repos & dummy point releases for all repos (incl wp & joomla)
 
4) patch distmaker to have a fallback pattern -e.g if not 4.6.5rc1 exists fall back to branch 4.6 & keep doing point versions.

My preference is for 4 but I realise it might not be seen as a generic enough need to go into BK

### DIFF
--- a/doc/cividist.md
+++ b/doc/cividist.md
@@ -6,12 +6,15 @@ CiviDist works by checking out the matching branch in each repo - so if you want
 exist in all repos.
 
 ## Setup: Make the web root
+```
 civibuild create dist --url http://dist.localhost ; 
+```
 
 ## Setup: Register your forks
  Note that if you don't fork one of these repos then you should still add a fork,
  but point it to the main civicrm repo so you have a consistent remote alias for all relevant repos.
 
+```
 cd build/dist/src
 git remote add myfork https://github.com/myfork/civicrm-core.git
 git config core.filemode false
@@ -31,15 +34,22 @@ git config core.filemode false
 cd build/dist/src/Wordpress
 git remote add myfork https://github.com/myfork/civicrm-wordpress.git
 git config core.filemode false
+```
 
 ## Periodic: Update code - this will retrieve from the remote alias - ie. myfork in the text below
+```
 cd build/dist
 env GIT_REMOTE=myfork cividist update 
+```
 
 ## Periodic: Build tarballs
+```
 cd build/dist
 cividist build myfork/4.6
+```
 
 ## Periodic: Cleanup old/orphaned tarballs
+```
 cd build/dist
 cividist prune
+```
--- a/doc/cividist.md
+++ b/doc/cividist.md
@@ -1,0 +1,45 @@
+## CiviDist
+Cividist generates tarballs from the git repos. If you wish to run it with your own repos you will need to do the 2
+setup steps first and then the last 3 when you want to build new tarballs (e.g nightly).
+
+CiviDist works by checking out the matching branch in each repo - so if you want to use a non-standard name it must 
+exist in all repos.
+
+## Setup: Make the web root
+civibuild create dist --url http://dist.localhost ; 
+
+## Setup: Register your forks
+ Note that if you don't fork one of these repos then you should still add a fork,
+ but point it to the main civicrm repo so you have a consistent remote alias for all relevant repos.
+
+cd build/dist/src
+git remote add myfork https://github.com/myfork/civicrm-core.git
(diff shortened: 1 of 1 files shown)


totten:
FWIW, I'm biased toward item 3.

 * In core, we make tags for new releases on all repos (even when there's no substantive change).
 * There have been several other scripts which assumed identical tag/branch names across repos (eg `cividist`, `releaser`, `givi`, and `drush.make.tmpl`).
 * If you ever need to look back and debug a past release, it's easier to do in git if you have the branches or tags. (No need to figure out which revisions got loaded via fallback.)
 * distmaker is already a bit complicated.

My vote would be to make a separate helper script to handle the branching -- call it something like `civi-auto-branch` -- which goes through each of the repos and makes a new branch (if it doesn't already exist).

```
## Usage
cd build/dist/src
civi-auto-branch origin/4.6 fuzion/4.6.5-fuzion

## Pseudocode
SRC_REMOTE=$(dirname $1)
SRC_BRANCH=$(basename $1)
TGT_REMOTE=$(dirname $2)
TGT_BRANCH=$(basename $2)

do_auto_branch "."         $SRC_REMOTE $SRC_BRANCH     $TGT_REMOTE $TGT_BRANCH
do_auto_branch "packages"  $SRC_REMOTE $SRC_BRANCH     $TGT_REMOTE $TGT_BRANCH
do_auto_branch "drupal"    $SRC_REMOTE 6.x-$SRC_BRANCH $TGT_REMOTE 6.x-$TGT_BRANCH
do_auto_branch "drupal"    $SRC_REMOTE 7.x-$SRC_BRANCH $TGT_REMOTE 7.x-$TGT_BRANCH
do_auto_branch "joomla"    $SRC_REMOTE $SRC_BRANCH     $TGT_REMOTE $TGT_BRANCH
do_auto_branch "wordpress" $SRC_REMOTE $SRC_BRANCH     $TGT_REMOTE $TGT_BRANCH

## ex: do_autofork <path> <from_remote> <from_branch> <to_remote> <to_branch>
function do_autofork() {
  pushd $path
    if [ ! -e .git/refs/remotes/$to_remote/$to_branch ]; then
      git branch $to_branch $from_remote/$from_branch 
      git push $to_remote $to_branch
    fi
  popd
}
```

totten:
On the new *.doc file -- great idea. It would be a bit easier to read in Github if you marked the example commands with triple-back-ticks, e.g.

```
echo "This is code!"
```

eileenmcnaughton:
like that?

Is your suggestion for 1 above also to handle it outside of the distmaker script?

totten:
Yup, the changes make it easier to read. :)

For renaming the tarballs, there is almost an option in `cividist` but it's not quite exposed properly. If you could change the value of `FILE_SUFFIX` to something else (eg blank or `nightly` or `fuzionnz`), then the values would be more stable. Currently, the assignment for `FILE_SUFFIX` is hardcoded to the date - but you could wrap it in a conditional (so you only set it to the date if there's no predefined value). Then run commands like:

```
env GIT_REMOTE=myfork FILE_SUFFIX=nightly civibuild build myfork/4.6
```

PR #163, opened 2015-07-25
migrate vagrant files from precise to trusty

patricklam:
I tried to use the existing Vagrant setup but it didn't work for me. This pull request migrates the Vagrant setup from precise32 (12.04 LTS) to trusty32 (14.04 LTS) and makes a few changes that are required due to the migration.

### DIFF
--- a/vagrant/trusty32-shared/Vagrantfile
+++ b/vagrant/trusty32-shared/Vagrantfile
@@ -1,0 +1,142 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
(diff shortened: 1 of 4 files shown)


totten:
Thanks, Patrick. I've been a little pressed right now, but I'll try to test/merge this week.

totten:
Confirmed that `vagrant up` works as expected (tested: trusty32-standalone), and `civibuild` inside the VM also appears to work correctly. The diff between `precise32-standalone` and `trust32-standalone` looks sensible.

Thanks again!

Issue #161, opened 2015-07-22
drupal_install has non-d8 command in it


eileenmcnaughton:
The drupal_install function is mostly generic enough to work with all 3 drupal versions - but there is one line it it which is not d8-compatible. I propose moving it to the specific install scripts

drush vset --yes file_private_path "${PRIVATE_ROOT}/${DRUPAL_SITE_DIR}"

eileenmcnaughton:
Actually @totten will check with you before I propose code - I'd have to move that line to a lot of places. Alternatively I could move this line

CIVI_UF="Drupal"

Higher up the individual files & drupal_install could do a check for it.

eileenmcnaughton:
refer branch https://github.com/eileenmcnaughton/civicrm-buildkit/tree/d8

eileenmcnaughton:
Also, note the command I'm using is 
 civibuild create bk-d8-fudev.co.nz --civi-ver master --type drupal8-demo --url http://bk-d8-fudev.co.nz

(more to keep my notes together on this ticket than anything else)

PR #165, opened 2015-07-28
D8 demo site 'working'

eileenmcnaughton:
This PR gets us to the point where a d8 demo can be spun up using build-kit - & I'd like to get it merged in / get a d8 demo firing on the civicrm.org site. However, there is one blocker to merging - which is 
#161 - the very last line in this PR affects other drupal versions.

Other than that the user add & role add & variable setting stuff is basically just commented out, but I don't think that is a blocker to changing the script so it starts to fire up.

 I'm hoping Torrance can look the role / user stuff since he'll fly though some of those fix-ups


totten:
I'm generally happy to merge things that will go into the new drupal8 folder, but these two bits should be fixed:

 * app/civibuild.conf should not be committed. it's local-only file. if not already, it should be in gitignore
 * probably want drupal8_uninstall to match drupal_uninstall.

Issue #171, opened 2015-08-21
Calling node on Debian sid fails because it's installed as "nodejs".


openbrian:
When this happens a simple simlink from /usr/local/bin/node to /usr/bin/nodejs works.

totten:
Yeah, it's annoying. You can't write scripts or docs which use the standard `node` command (because that breaks on Debian), and you can't write scripts or docs which use Debian's `nodejs` command (because that breaks everywhere else). 

I think the best prospect would be to patch [`civi-download-tools`](https://github.com/civicrm/civicrm-buildkit/blob/master/bin/civi-download-tools#L261) to explicitly check for the Debian situation and either:

 * Autocreate a symlink from `buildkit/bin/node` to `/usr/bin/nodejs`
 * Encourage the user to install `nodejs-legacy`. (Which is a bad name... but whatever.)

totten:
@openbrian , I've merged https://github.com/civicrm/civicrm-buildkit/pull/173 . I didn't test a full/clean install, but it did work in quick/hacky testing.

PR #173, opened 2015-08-24
civi-download-tools - Workaround for Debian `nodejs` binary (#171)

totten:


openbrian:
Nice, 3 birds with 1 stone.

PR #176, opened 2015-08-25
Add Drupal 8 user, civicrm web role and and whole bunch of permissions.

torrance:


totten:
FWIW, calling drush to add each permission individually was really slow. (At least with D7... you have to bootstrap drush+Drupal each time.) Any chance you could still use "$PRJDIR/src/drush/perm.php" (or a copy/fork thereof) to set them en masse?

PR #182, opened 2015-09-16
Default to latest version of WP

colemanw:


app/config/wp-demo/download.sh
```diff
@@ -4,12 +4,12 @@
 
 ###############################################################################
 
-[ -z "$CMS_VERSION" ] && CMS_VERSION=4.3
+[ -n "$CMS_VERSION" ] && CMS_VERSION=" --version=$CMS_VERSION"
```
totten: The content `4.3` and `--version=4.3` should be in different variables. It hurts my brain when the structure/purpose of a variable changes.

What about:

```php
WPCLI_ARGS=
[ -n "$CMS_VERSION" ] && WPCLI_ARGS="$WPCLI_ARGS --version=$CMS_VERSION"
```

totten: It may be a little nitpicky, but I care b/c `CMS_VERSION` is a document variable in https://github.com/civicrm/civicrm-buildkit/blob/master/src/civibuild.defaults.sh#L86 . Someone skimming the list of vars wouldn't expect it to include `--version=4.3`.

PR #188, opened 2015-09-29
app/config/* - Don't set extensionsDir on v4.7+

totten:
The core app should handle defaults better.

### DIFF
--- a/app/config/doctrine/install.sh
+++ b/app/config/doctrine/install.sh
@@ -22,9 +22,13 @@ CIVI_CORE="${WEB_ROOT}/sites/all/modules/civicrm"
 CIVI_SETTINGS="${WEB_ROOT}/sites/${DRUPAL_SITE_DIR}/civicrm.settings.php"
 CIVI_FILES="${WEB_ROOT}/sites/${DRUPAL_SITE_DIR}/files/civicrm"
 CIVI_TEMPLATEC="${CIVI_FILES}/templates_c"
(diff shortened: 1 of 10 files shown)


totten:
Depends: https://github.com/civicrm/civicrm-core/pull/6845

totten:
Rebased and resolved conflicts in `wp-demo`. Also, I found that WP 4.4 was having problems, so updated `wp-cli` and `app/config/wp-demo` accordingly.

kcristiano:
@totten Can you share the problems with 4.4?   I've been OK in production (not buildkit) but if there is an issue, I'd like to take a look.

totten:
Thanks, @kcristiano. I think the problems were specific to buildkit:

1. `wp core install` (or maybe `wp core download`?) was throwing an error like `Failed opening required '...wp-includes/embed-functions.php'` Some quick googling suggested that the files were changed circa WP 4.4. Upgrading to the latest `wp-cli` resolved it. (Buildkit had v0.17; the latest was v0.21.1.)

2. The `twentythirteen` theme (which is used by `wp-demo/install.sh`) wasn't activating. Changed to download+activate.

Tangentially, I touched some lines that setup the path variables. That's not because anything was wrong with the logic -- but there was a merge-conflict.

kcristiano:
Makes sense @totten and yes, wp-cli has to be at least 20.4 to work with 4.3.1 and 4.4.   I forgot we were using twenythirteen, that was unbundled with 4.4.   Plus your regex is way better than what I used.

PR #197, opened 2015-11-25
CRM-17609 - install-dashboard.php - Use new user-sync signature (4.7+)

totten:


app/config/drupal-demo/install-dashboard.php
```diff
@@ -7,7 +7,14 @@
...
 }
 $users = explode(';', getenv('INSTALL_DASHBOARD_USERS'));
-CRM_Core_BAO_CMSUser::synchronize(FALSE);
+
+// WISHLIST: CMSUser::synchronize probably merits an API but lacks test coverage.
+if (is_callable(array('CRM_Core_BAO_CMSUser', 'synchronize'))) {
+  CRM_Core_BAO_CMSUser::synchronize(FALSE); // 4.6 and earlier
+}
+else {
+  CRM_Core_Config::singleton()->userSystem->synchronizeUsers(); // v4.7+
```
colemanw: As of https://github.com/civicrm/civicrm-core/commit/091db9085e779dc6b50f1644c721846437b6c844 the more concise way to write this would be `CRM_Utils_System::synchronizeUsers()`

totten: Ah, ok. That commit is -100 sloc, so it must be good. :)

totten: See https://github.com/civicrm/civicrm-buildkit/pull/198

PR #201, opened 2015-12-09
added civirules to the drupal demo sites

jaapjansma:
Added civirules to the buildkit


totten:
Nice -- thanks for the PR! Merged.

PR #202, opened 2015-12-16
civi-download-tools - Download drush8

totten:
Drush8 is distributed as a PHAR, so it seems fairly easy now to have multiple copies of drush available. This adds a command `drush8` to buildkit's `bin/`.

This is necessary because we need to support both PHP 5.3 (for our test env and D7; using an older drush) and PHP 5.5 (for D8; using newer drush).


eileenmcnaughton:
cool

PR #207, opened 2016-01-07
adjust phpcs and jshint path 

mallezie:
I've adjusted the phpcs and jshint path in the civilint script to allow civilint when the builkit is not defined in the path. 

This way i can run ~/buildkit/bin/civilint


totten:
Thanks! I didn't see this come through when it was first submitted, but it looks good.

PR #213, opened 2016-01-31
update composer.json to get latest version of wp-cli.

kcristiano:


totten:
OK. We should aso update `composer.lock`.

kcristiano:
Closing will issue new PR

PR #217, opened 2016-02-03
Can we grab in the latest version of AMP?

eileenmcnaughton:


totten:
Good idea, although this PR has some problems. (It doesn't update `composer.lock` and has a merge-conflict). I'll submit a patch which downloads `amp` as a PHAR.

PR #220, opened 2016-02-05
Download civix from github, not sourceforge

ejegg:
Reduces the number of external sites we rely on.
Uses the 'release' script to generate the same tarball.

### DIFF
--- a/bin/civi-download-tools
+++ b/bin/civi-download-tools
@@ -19,7 +19,8 @@ TMPDIR="$PRJDIR/app/tmp"
 LOCKFILE="$TMPDIR/civi-download-tools.lock"
 LOCKTIMEOUT=90
 CIVIXVER=15.04.1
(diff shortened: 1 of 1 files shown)


totten:
Ah, interesting idea. So it looks like this patch calls [civix/releaser](https://raw.githubusercontent.com/totten/civix/master/release) which:
 * Clones the repo
 * Runs `composer install` (which downloads ~16 packages)
 * Deletes unnecessary files; puts the rest in a tarball.

For some background, civix was originally only available through composer. (e.g. "git clone https://.../civix && composer install") We had periodic download problems (eg upstream rearranging repos; composer needing an update; Github outage), so then switched to building/publishing tarballs (by way of Sourceforge).

My impression is that `composer install` (~16 packages) has more systemic risk than any single download provider.

I'm happy to post the tarballs to different server. How about changing the download URL to `https://download.civicrm.org/civix/civix-15.04.1.tar.bz2` ?

totten:
There's discussion about this in the other thread. Seems we're agreed that this formulation has some problems. Closing in anticipation of a different approach.

PR #225, opened 2016-02-15
add missing dependency, angular profiles extension, plus clean-up.

ginkgomzd:


ginkgomzd:
@GinkgoFJG

totten:
Seems good in terms of loading the new dependency. Since there are now multiple builds using these (drupal-demo and drupal8-demo), it makes sense to declare the caches in a more central spot.

I do get an error when navigating to `Events => Manage Events => Rain-forest Cup => Configure => Volunteers`, but that's probably unrelated. (HTTP 500 on the AJAX request for `civicrm/event/manage/volunteer?reset=1&snippet=json&force=1&id=3&action=update&component=event`)

PR #229, opened 2016-02-19
add some useful commands

ginkgomzd:
adds these commands

list
the --full option to show
edit

### DIFF
--- a/bin/civibuild
+++ b/bin/civibuild
@@ -235,34 +235,18 @@ function civibuild_show() {
   fi
 
   if [ -n "$SHOW_FULL_BUILD_CONF" ]; then
-    civibuild_show_full_config
+    civibuild_show_summary $PERSISTENT_VARS
   else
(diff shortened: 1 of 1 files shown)


totten:
Sounds useful.

How would you feel about adding it as a subcommand `civibuild ls` or `civibuild list`:

https://github.com/civicrm/civicrm-buildkit/blob/master/bin/civibuild#L502

To get the path to the build dir, you could use `$BLDDIR` instead of `$(dirname `which civibuild` 2>/dev/null)'/../build'`. This would have the added benefit of working in some custom configurations where BLDDIR has been customized.

ginkgomzd:
adds these commands
 - list
 - the --full option to show
 - edit

ginkgomzd:
@totten This is ready again for your review.
Thanks

totten:
Superceded by #233.

PR #231, opened 2016-02-22
Parse options

ginkgomzd:
branched from
https://github.com/civicrm/civicrm-buildkit/pull/229

Add validation of ACTION  and remove restriction on order of arguments.
besides <action> <site-name>[/<ms-id>], arguments may be in any order.


totten:
Superceded by #233.

PR #233, opened 2016-02-22
Smart SITE_NAME

ginkgomzd:
branched from https://github.com/civicrm/civicrm-buildkit/pull/231

if SITE_NAME is not passed and current working directory is below a CMS root, or above only one build, then set SITE_NAME intellegently.

### DIFF
--- a/bin/civibuild
+++ b/bin/civibuild
@@ -235,34 +235,18 @@ function civibuild_show() {
   fi
 
   if [ -n "$SHOW_FULL_BUILD_CONF" ]; then
-    civibuild_show_full_config
+    civibuild_show_summary $PERSISTENT_VARS
   else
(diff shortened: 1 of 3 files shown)


totten:
The autodetection is really cool! That's been my biggest annoyance.

I made a few changes which will be pushed along:
 * 7e0ebc348ab - Consolidate `ls | list | builds` to just `list`. `list` is not hard on the fingers.
 * 28f1c9a7ac7 - Simplify search logic. Recursive search is unnecessary+slow.

For `civibuild edit` - it works, and I've got no objection to it, although I don't really understand it either. Anytime I felt a temptation to edit those files directly, it's also felt safer to blow-away the build. But there may be good reasons for it.

Issue #234, opened 2016-02-29
civibuild keeps showing help


AbdealiJK:
I'm not quite sure if this is a docs issue or a civibuild issue.

I installed buildkit and now tried to run `civibuild` using the command [here](https://github.com/civicrm/civicrm-buildkit/blob/master/doc/civibuild.md) and it just keeps showing me the `-h` info again and again.
I'm running: `$ civibuild create dmaster --url http://localhost:8001 --admin-pass s3cr3t`

```
ajk@ajk-ThinkPad-R60:~/Documents/civi$ civibuild create dmaster --url http://dmaster.localhost --admin-pass 111
Common options:
  <build-name>        The name of the sub-directory to build
  <ms-id>             The unique id for a site within a multisite build

Syntax: civibuild create    <build-name>[/<ms-id>] [options]
Syntax: civibuild download  <build-name>           [options]
Syntax: civibuild install   <build-name>[/<ms-id>] [options]
Syntax: civibuild reinstall <build-name>[/<ms-id>] [options]
Description: Download and/or install the application
  --type <type>       The name of the download/install scripts. (If omitted, assume <build-name>) [Optional]
  --web-root <path>   The full path to the website root. [Default: /home/ajk/buildkit/build/<build-name>]
  --civi-ver <ver>    The branch or tag of CiviCRM desired (master, 4.4, 4.3, 4.3.0, etc) [Optional]
  --cms-ver <ver>     The release of the CMS desired [Optional]

  --url <url>         The public URL of the site
  --title <title>     The title of the site
  --admin-user        Name of the CMS's administrative user
  --admin-pass        Password for the CMS's administrative user
  --admin-email       Email of the CMS's administrative user
  --demo-user         Name of the CMS's demo user
  --demo-pass         Password for the CMS's demo user
  --demo-email        Email of the CMS's demo user

  --force             If necessary, destroy pre-existing files/directories/DBs
                      (For "reinstall", "--force" is implicit.)

Syntax: civibuild list
Description: Display a list of build-names found in the build directory. 

Syntax: civibuild edit <build-name>[/<ms-id>]
Description: Edit the <build-dir>/<build-name>[.<ms-id>].sh config file

Syntax: civibuild show <build-name>[/<ms-id>] [options]
Description: Show key details about the build
  --html <dir>        A new HTML dir which will report build status in detail
  --new-scan <file>   A new JSON file which summarizes the git structure
[...]

PR #235, opened 2016-03-01
bin/civibuild - civibuild_validate_site_name is too strict (#234)

totten:
When creating a new build, the build doesn't exist yet! Therefore, it fails
the validation rule. The validation rule is overzealous.

@ginkgomzd, is there a reason why the validation needed to check if the folder existed?
I did a quick spot-check of a few scenarios, and they seem to work without
the check:

 * `civibuild create dmaster` (under buildkit base dir)
 * `civibuild restore dmaster` (under buildkit base dir)
 * `civibuild restore` (under the dmaster folder -- autodetect the build)
 * `civibuild restore d46` (under the dmaster folder -- use explicit input, no autodetect)

### DIFF
--- a/bin/civibuild
+++ b/bin/civibuild
@@ -61,7 +61,7 @@ Description: Download and/or install the application
                       (For "reinstall", "--force" is implicit.)
 
 Syntax: $APP list
-Description: Display a list of build-names found in the build directory. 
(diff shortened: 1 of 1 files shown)


AbdealiJK:
:D I was just about to ask why this wasn't merged yet

PR #241, opened 2016-04-05
civibuild - Split command files. Add URL_TEMPLATE and more tests.

totten:


tests/phpunit/CivibuildGeneralTest.php
```diff
@@ -0,0 +1,164 @@
...
+
+    // FIXME: Bug!
+    //    $resultInDir = ProcessUtil::runOk(
+    //      $this->cmd("civibuild $command")->setWorkingDirectory($buildDir)
+    //    );
+    //    $this->assertContains("[[Detected SITE_NAME: {$this->buildName}]]\n", $resultInSubdir->getOutput());
+    //    $this->assertEquals(
+    //      $resultWithArg->getOutput(),
+    //      str_replace("[[Detected SITE_NAME: {$this->buildName}]]\n", "", $resultInDir->getOutput())
+    //    );
```
totten: @ginkgomzd, while adding unit tests for civibuild, I bumped into an edge-case where `civibuild_detect_site_name` stumbles -- it works correctly in subdirs (e.g. `build/dmaster/sites/all/modules`) but not in root dirs (e.g. `build/dmaster`). The example above demonstrates this.

PR #242, opened 2016-04-14
Update wp-cli version for 4.5 compatibility

kcristiano:
wp-cli 0.23 required for WP 4.5


totten:
Thanks. 

Confirmed that `wpmaster` still builds locally. Also skimmed http://wp-cli.org/blog/version-0.23.0.html and didn't see anything scary.

Merged.

PR #243, opened 2016-04-19
Infra 178

yashodha:


totten:
@yashodha , cool!

FWIW, I'm getting an error like this:

```
STDERR > +++ wp civicrm api extension.install key=org.civicrm.volunteer debug=1
STDOUT > Array
(
    [is_error] => 1
    [error_message] => Unknown extension: org.civicrm.volunteer
)
```

It probably needs another patch to download the code. Note that `app/config/wp-demo` has two main scripts, `download.sh` and `install.sh`. In particular, `download.sh` calls `git clone` for each dependency:

https://github.com/civicrm/civicrm-buildkit/blob/master/app/config/wp-demo/download.sh 

...so maybe just add something like:

```
## At the top
[ -z "$VOL_VERSION" ] && VOL_VERSION='4.4-1.x'

## Toward bottom, with the other clone statements
git clone ${CACHE_DIR}/civicrm/civivolunteer.git -b "$VOL_VERSION" civicrm/civicrm/tools/extensions/civivolunteer
```

Also, the different versions of Civi should be matched up with different versions of CiviVolunteer. Note how the `d45`, `d46`, `dmaster`, etc have `VOL_VERSION` specified. The same thing would make sense for `wp46`, `wpmaster`, etc.

https://github.com/civicrm/civicrm-buildkit/blob/master/src/civibuild.aliases.sh

yashodha:
Thanks @totten . Implemented the same in #244

PR #247, opened 2016-04-27
PCHR-894: Fix concurrency issue in drush make

AkA84:
The `drush make` command, when used to build a project with many dependencies (ours has around 50), has a concurrency issue ([details](https://github.com/drush-ops/drush/issues/985)) in the [`drush_make_process`](https://github.com/drush-ops/drush/blob/master/commands/make/make.drush.inc#L710) function, where the projects array content might not make it into the function, resulting in the dependencies not getting downloaded.

By increasing the concurrency value (the default value is 1) the problem seems to be solved, the issue could not be replicated anymore

### DIFF
--- a/app/config/hr15/download.sh
+++ b/app/config/hr15/download.sh
@@ -19,4 +19,4 @@ cat "$SITE_CONFIG_DIR/drush.make.tmpl" \
   | sed "s;%%HR_VERSION%%;${HR_VERSION};" \
   > "$MAKEFILE"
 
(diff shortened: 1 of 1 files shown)


totten:
Fascinating. Aside: does it have any perceptible influence on performance?

AkA84:
Not that I have noticed! This issue was pretty random, sometimes I could run the `civibuild create` 10+ times without ever encountering it, sometimes I could get it 5 times in a row :/ I even read in another issue that this might be related to the OS's resources at the time the command is ran, pretty tricky

PR #248, opened 2016-05-02
Update wmf config in buildkit to match latest wmf version 

eileenmcnaughton:
Since it seems we ship a wmf buildkit config it makes sense that it should be the latest wmf_config


eileenmcnaughton:
@totten  are you happy for me to 'merge at will' PRs that ONLY touch the wmff folder within buildkit?

totten:
@eileenmcnaughton yup, that sounds good.

PR #250, opened 2016-05-16
PCHR-655: patch to sort out pagination issue with views_merge_rows

nishant-bhorodia:
**Issue:**
On staff directory, even if you choose different number of items to show per page, that number seems to be ignored. (e.g. select 10 per page and get 3 result on the first page and 5 result next page.

**Reason:**
It was happening because views_merge_rows was not able to handle pagination correctly. The patch sorts out that issue.


totten:
Accepted. Patch only affects CiviHR build and submitted by a CiviHR maintainer.

PR #251, opened 2016-05-17
civibuild - Add option to create Backdrop/Drupal/WordPress site with Civi tarballs

totten:
To facilitate testing the graphical installation process using tarballs, this
allows you to create a new Backdrop/Drupal/WordPress site *without* git; instead,
it just extracts the given tarball.

For example, suppose you want to manually test the installation of a
nightly tarball on an empty Drupal site. Run:

```
civibuild create foo --type drupal-empty --dl sites/all/modules=http://dist.civicrm.org/by-date/latest/master/civicrm-4.7.8-drupal-20160516.tar.gz
```

Or suppose you've just run distmaker and produced an experimental zip
archive for WordPress installations.  You can try it out on a new/empty
WordPress site by running:

```
civibuild create foo --type wp-empty --dl wp-content/plugins=$PWD/civicrm-4.7.8-wordpress.zip
```

Tangentially, you can also download other tarballs and zipfiles with the
same syntax. Just add multiple `--dl` parameters.


totten:
For the somewhat common case of testing installation across all CMSs, I've added another command, `civihydra`, which can take a list of tarballs (local files or HTTP URLs) and setup empty sites for each of them. More docs are included in online help and commit notes.

PR #256, opened 2016-05-26
allow to pass civihr version as civibuild parameter

omarabuhussein:
Its important for us at @compucorp to be able pass civihr version as civibuild parameter specially when creating multiple civihr websites ( instances ) and since civihr is still part of  buildkit I think it make sense to submit this PR here.

### DIFF
--- a/src/civibuild.parse.sh
+++ b/src/civibuild.parse.sh
@@ -261,6 +261,11 @@ function civibuild_parse() {
         shift
         ;;
 
+      --hr-ver)
+        HR_VERSION="$1"
+        shift
+        ;;
+
       *)
         if [ "${OPTION::1}" == "-" ]; then
           echo "Unrecognized option: $OPTION"


totten:
This is should be safe to merge and seems reasonable given CiviHR's unique structure, although I don't *think* it's really necessary, and I don't want to set a precedent where folks add options for lots of different extensions.

In terms of why it's not necessary... the `civibuild` scripts are mostly working with environment variables. When calling civibuild, you can usually set these inline using the `env` command, e.g. compare

```bash
civibuild create hr16
env HR_VERSION=9.8.7 civibuild create hr16
env HR_VERSION=9.8.7 VOL_VERSION=6.5.4 OFFLINE=1 civibuild create hr16
```

(And there are other ways to manage default values for env variables, like `~/.profile` or `app/civibuild.conf`.)

PR #257, opened 2016-06-05
Update README.md

stevekessler:
Added link to Vagrant Virtual Machine from Vail Sprint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@ full collection -- installing each individually takes a lot of work.
 This is the same collection of tools which manages the test/demo/release
 infrastructure for civicrm.org.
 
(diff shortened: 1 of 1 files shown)


totten:
I like the idea of referencing the vagrant box, but it would be more consistent to list it as another download option, e.g.

 * Download: Option #1: Vagrant
 * Download: Option #2: Full Stack Ubuntu
 * Download: Option #3: Other Systems
 * Download: Option #4: Upgrade

totten:
@stevekessler, I've updated the README to present Vagrant as one of the download options. Also, the list of 'System Requirements' is less important if you're using the `civi-download-tools --full` or `vagrant up`, so I've pushed that down.

PR #249, opened 2016-05-13
Change hr16 build to checkout by tag

omarabuhussein:
When we create a new **hr16** build with this command : 

`civibuild create hr16 --civi-ver 4.7 --url http://localhost:8900` , It doesn't work   because **civicrm-drupal** repo doesn't have a branch named : **7.x-4.7** , the only one related is **7.x-4.7.7-rc** unlike previous civicrm versions where you can find 7.x-4.6, 7.x-4.5 ... etc . Also there no branches in **civicrm** or **civicrm-packages** repos  named 4.7 , the only ones related are named 4.7.7-rc .

also I changed all the repos to get checked out by tag not by branch name to insure stability since when we tried to checkout the latest civicrm stable version which is 4.7.7 like this :
`civibuild create hr16 --civi-ver 4.7.7 --url http://localhost:8900`
The following error was thrown :

`WD php: CRM_Core_Exception: Invalid Entity Filter in CRM_Core_BAO_CustomGroup::validateSubTypeByEntity()     [error]
(line 658 of
/Users/testuser/GitHub/buildkit/build/hr16/sites/all/modules/civicrm/CRM/Core/BAO/CustomGroup.php).
Cannot modify header information - headers already sent by (output started at                                [warning]
/Users/testuser/GitHub/buildkit/vendor/drush/drush/includes/output.inc:38) bootstrap.inc:1232
CRM_Core_Exception: Invalid Entity Filter in CRM_Core_BAO_CustomGroup::validateSubTypeByEntity() (line 658 of /Users/testuser/GitHub/buildkit/build/hr16/sites/all/modules/civicrm/CRM/Core/BAO/CustomGroup.php).
Drush command terminated abnormally due to an unrecoverable error.                                           [error]` 

but when trying to create a build with  4.7.6 :

`civibuild create hr16 --civi-ver 4.7.6 --url http://localhost:8900`

everything was working perfectly

PR #262, opened 2016-07-01
(Don't merge yet) Allow to checkout repos by bransh or tag

omarabuhussein:
Its really useful sometimes to create a new civicrm build via the buildkit using a specific tag instead of branch , This PR add a new option called **checkout** which allow you to specify how the repositories going to be checked out either by tag or branch name .

Examples :

create civicrm drupal demo instance using civicrm 4.7.7 release
```
civibuild d4777 --type drupal-demo --civi-ver 4.7.7 --checkout tag -url http://localhost:8135
```

create civicrm drupal demo instance using latest master branch
```
civibuild dmaster --type drupal-demo --civi-ver master --checkout branch -url http://localhost:8135
```

create civicrm drupal demo instance using 4.7 branch
```
civibuild d47 --type drupal-demo --civi-ver 4.7 --checkout branch -url http://localhost:8135
```

This option for now work for the following build types :

- drupal-demo
- hr15
- hr16

### DIFF
--- a/app/config/drupal-demo/drush.make.tmpl
+++ b/app/config/drupal-demo/drush.make.tmpl
@@ -26,14 +26,14 @@ libraries[civicrmdrupal][destination] = modules
 libraries[civicrmdrupal][directory_name] = civicrm/drupal
 libraries[civicrmdrupal][download][type] = git
 libraries[civicrmdrupal][download][url] = %%CACHE_DIR%%/civicrm/civicrm-drupal.git
-libraries[civicrmdrupal][download][branch] = 7.x-%%CIVI_VERSION%%
+libraries[civicrmdrupal][download][%%CHECKOUT%%] = 7.x-%%CIVI_VERSION%%
 libraries[civicrmdrupal][overwrite] = TRUE
 
 libraries[civicrmpackages][destination] = modules
 libraries[civicrmpackages][directory_name] = civicrm/packages
 libraries[civicrmpackages][download][type] = git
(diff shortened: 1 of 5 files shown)


omarabuhussein:
@totten @AkA84 

What do you think about this ?

AkA84:
@omarabuhussein to me it looks great, it makes the tool more flexible. Only thing would be to make sure the `--checkout` param can be only `branch` or `tag`, without accepting some other value

totten:
@omarabuhussein, I went along with similar (but narrower/less flexible) patch when it only affected HR builds... but don't really understand the use-case where `drush.make` needs to differentiate between a branch or a tag. For example, the current file `app/config/drupal-clean/drush.make.tmpl` uses `branch`:

```
libraries[civicrm][download][branch] = %%CIVI_VERSION%%
```

But it still lets me use tags for CIVI_VERSION:

```
$ civibuild create delme --type drupal-clean --civi-ver 4.7.5
$ cd delme/sites/all/modules/civicrm
$ git status
HEAD detached at 4.7.5
nothing to commit, working directory clean
```

What's the problem being solved by changing this `tag`?

(In case it's relevant, my local system is OSX 10.10 running git version 2.6.4.)

omarabuhussein:
well... I wasn't aware that keeping it like this :
`libraries[civicrm][download][branch] = %%CIVI_VERSION%%`
will accept tag names too and not branches only. But knowing this I now wonder if this PR https://github.com/civicrm/civicrm-buildkit/pull/249  was even necessary .

PR #263, opened 2016-07-06
(CiviHR) Improve buildkit install process

AkA84:
We improved the install process of CiviHR, finally automating things that up to this point were done manually.

While I was at it, I refactored the `install.sh` files and bumped the `CIVI_VERSION` of `hr16` to 4.7.8


AkA84:
@totten ping

AkA84:
@totten could we have this merged? It's quite important for us, it doesn't affect buildkit as a whole, just our build type

totten:
You got it!

PR #265, opened 2016-07-14
civibuild.md - improved quick-start instructions

bdombro:


GinkgoFJG:
I'm not sure exactly what problems you're trying to solve, but here's some general feedback:

* The installation of buildkit is already addressed in the repo's README. Rather than duplicate that here, let's add a link back to the main README.
* The vim stuff probably doesn't belong here. That sounds like a personal preference. I don't use vim, for example.
* I, too, have noticed some filesystem issues following a buildkit install, but I haven't taken the time to document them. If I recall correctly, I create the sites as me, resulting in the web server (running under a different user) being unable to write certain places. Having a problem with writing to cache sounds familiar. I would classify this as a bug, however. Rather than add it to the doc, we should open an issue. Buildkit is already doing a lot to manage filesystem permissions -- we might just be missing this case.
* This is not documented anywhere: To specify which version of CiviVolunteer to install, you have to set an environment variable VOL_VERSION before running your buildkit command. If none is set, buildkit falls back to some default, which appears to be inappropriate. Maybe we should update that. Also, we recommend users of CiviCRM 4.6 and 4.7 run the latest 2.1.x of CiviVolunteer, not 1.4.

ginkgomzd:
re VOL_VERSION defaults...
Looks like VOL_VERSION is indeed out of date.
One option is to remove CiviVolunteer from the demo.... Just pointing that out. I believe someone could install and enable the extension on demo.civicrm.org anyway? And we want to be directing people to try out volunteer on http://civivolunteer.ginkgostreet.com/. This should be noted on the demo.civicrm.org landing page.

If we just want to update the default config, we need to update all of the download.sh files in all of the build configs. e.g. https://github.com/civicrm/civicrm-buildkit/blob/master/app/config/drupal-demo/download.sh#L8

I see drupal8-demo, wp-empty, drupal-demo, wp-demo have VOL_VERSION defined.

GinkgoFJG:
I don't see why we should remove it from the demo. I imagine c.o's rebuild script must define the VOL_VERSION, so fixing this is for everyone else's benefit.

And no, extension installation is locked down on the public demo.

totten:
* +1 for putting a link back to the buildkit README for general download instructions.
     * The main README page is getting a bit long, so I can sympathize with wanting a shorter version to include here. But those three download options are really very distinct. Perhaps we should edit the main README to just show links, e.g. https://github.com/totten/civicrm-buildkit/tree/master-readme ?
 * It shouldn't be necessary to set permissions manually. When running `amp config`, it asks about file permissioning (www username; whether to use linux ACLs; etc). Then, when creating any new builds, it calls `amp datadir sites/default/files` to set the permissions accordingly.
    * Aside: "It shouldn't be necessary" != "it never happens". For example:
      * A traditional 'world writeable' policy is pretty flaky; if you use it and frequently switch between WWW/CLI, then you'll need to fix permissions regularly. But if you have ACL's or if httpd runs as your regular user, then it should be automatic.
      * There could be bugs or unexpected use-cases, in which case... we want to know about those...
 * I had to do a double-take on the EDITOR. Personally, I tend to `apt-get install joe`  which sometimes has the unintended side-effect of changing the system default... so I was worried that maybe this habit slipped into builtkit somehow... but, to my reading, buildkit remains agnostic to the EDITOR. As long as buildkit preserves/respects the system default, I don't think it's our job to teach people how to change their EDITOR.
 * If you know a few people who are specifically working on new VMs + buildkit + CiviVolunteer, it could be really useful to give them a custom/tuned README. However, for the mainline branch, we need to anticipate a more diverse audience.

ginkgomzd:
@GinkgoFJG  It sounds like you also don't see why it should stay in the demo.
Sorry, I thought it was obvious the reasons why removing it is being considered:
 * Situations like this where the need for update was overlooked in the release process
 * Wanting to promote the demo we host and reviewers having a consistent demo experience
 * To some uncertain degree, how much the new promotion of off-site demo's on the demo.civicrm.org landing page, tips the scale in either direction. Mentioned on the partners' mattermost channel (https://chat.civicrm.org/civicrm/pl/157osr4k4ffcdr4rxguqyhp75r)

Reasons off the top of my head to keep it in the demo:
 * People expecting it to be here might not know where it went
 * More demo's get more reviewers
 * It's deemed strategic to the CiviCRM demo over-all

bdombro:
Thanks for all the feedback and ideas. To give everyone a little context, I was approaching this from the angle that bug-fixes and major changes can take a lot of time, and it would be nice to have full end-to-end instructions for getting started on Ubuntu 14.04 with the current code, and then these instructions can be updated as the bugs/changes occur.
I was inspired to contribute these instructions because it took significant effort/expertise to get up and running with the current instructions.

On Thu, Jul 14, 2016 at 4:04 PM, Michael Z Daryabeygi <notifications@github.com> wrote:
@GinkgoFJG [https://github.com/GinkgoFJG] It sounds like you also don't see why it should stay in the demo.
Sorry, I thought it was obvious the reasons why removing it is being considered:

 * Situations like this where the need for update was overlooked in the release process
 * Wanting to promote the demo we host and reviewers having a consistent demo experience
 * To some uncertain degree, how much the new promotion of off-site demo's on the demo.civicrm.org landing page, tips the scale in either direction. Mentioned on the partners' mattermost channel ( https://chat.civicrm.org/civicrm/pl/157osr4k4ffcdr4rxguqyhp75r [https://chat.civicrm.org/civicrm/pl/157osr4k4ffcdr4rxguqyhp75r] )

Reasons off the top of my head to keep it in the demo:

 * People expecting it to be here might not know where it went
 * More demo's get more reviewers
 * It's deemed strategic to the CiviCRM demo over-all

totten:
Re: Volunteer in demo - I think it's OK to keep it in or move it to a separate demo system. But note that you cannot download arbitrary extensions in the current demo infra -- because that would essentially allow anonymous/untrusted people to run arbitrary code. I don't think it's a good idea to allow demo users to download extensions (unless we have properly sandboxed, per-person, ephemeral demo sites).

totten:
@bdombro , generally kudos on the installation! Getting a working development build is a significant task! And thanks for taking initiative to update this.

Based on some of the lessons from #265, I'm opening #266 and closing #265. A few final thoughts on why:
 * The phrase "Quickstart" is no longer accurate. IIRC, past versions of `README.md` actually had a quickstart (more like #265), but it kept growing more complicated as we tried harder and harder to make it easy. (Take away: it's really hard to make this stuff "easy".)
 * At this point, `civibuild.md` is more like "the reference manual for `civibuild`". It's not intended to be the first page a new contributor sees -- just one part of the process. Something like [README.md](README.md) or https://github.com/civicrm/civicrm-dev-docs/ should be the starting point.
 * There's another document, `daily-coding.md`, which talks more about things like updating code. 

@GinkgoFJG @ginkgomzd - In terms of keeping the demo build up-to-date (without having it use bleeding-edge code)... you might want to create a checklist of "things to do when pushing a new release of CiviVolunteer". The list would include things like making git tags, testing the upgrade, updating `src/civibuild.aliases.sh`, posting to Twitter, sending emails, etal. For comparison, CiviCRM-core has a checklist at http://wiki.civicrm.org/confluence/display/CRM/Release+checklist .

PR #268, opened 2016-07-25
Update radix_theme version in hr16

davialexandre:
A recent update in Drupal 7.50 changed the behavior drupal_get_filename to
throw an error when a file is missing. This caused some errors in the SSP,
because the old rc version of the theme was trying to handle files from
the panapoly_admin module without checking first if the module was installed.


totten:
Thanks. Didn't initially recognize you were working with CiviHR project. I usually to CiviHR team on any changes for `hr16`, etal.

PR #270, opened 2016-08-01
Enable username_to_external_id module during hr15+hr16 installation

omarabuhussein:
This affect only hr16 and hr15 build types (civihr) . 

I enable username_to_external_id module which is shipped with civihr and used to fill the contanct external_id field with it's drupal username if exists.


totten:
@omarabuhussein , OK I'll wait on merging. Just ping `@totten` when it's ready.

omarabuhussein:
@totten  can you please merge this... thanks

PR #271, opened 2016-08-05
PCHR-959: Update webform_civicrm version

AkA84:
@colemanw just [released](https://github.com/colemanw/webform_civicrm/pull/43#issuecomment-237612031) a new stable version (4.16) of the [webform_civicrm](https://www.drupal.org/project/webform_civicrm) module, which contains a couple of patches that we submitted over the course of these past months, so we can drop the dependency to the dev version of the module instead

### DIFF
--- a/app/config/doctrine/drush.make.tmpl
+++ b/app/config/doctrine/drush.make.tmpl
@@ -72,10 +72,10 @@ projects[redirect][subdir] = contrib
 projects[redirect][version] = 1.0-rc1
 
 projects[webform][subdir] = contrib
-projects[webform][version] = 3.19
+projects[webform][version] = 4.12
 
 projects[webform_civicrm][subdir] = contrib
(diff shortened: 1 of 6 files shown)


colemanw:
Thnks for this @AkA84.
It looks like there are a few other places that could also use a version bump:
https://github.com/civicrm/civicrm-buildkit/search?utf8=%E2%9C%93&q=webform_civicrm

colemanw:
Some of those will also require updating the `webform` module version as well.

AkA84:
Hi @colemanw i've updated `webform` and `webform_civicrm` to the latest versions (the former had to be upgraded as well to avoid dependencies issues) also in the other build types you pointed out. Let me know if it's ok, thanks

colemanw:
Thanks @AkA84 :)

PR #273, opened 2016-08-23
Exclude DAO,examples and tools directories via question mark regex in clivilint

omarabuhussein:
I changed the regex in **civilint** that exclude the **DAO,examples and tools** directories to use the question mark instead of using the absolute path .

It should still work for civicrm , and it helpful in our work with civihr since civihr is installed currently in **civicrm/tools** directory so it always get excluded by civilint.


totten:
Thanks

PR #277, opened 2016-10-14
PCHR-1495: Apply available drupal updates

coldrunKacper:
- Updaing versions of Drupal modules: jQuery Update, Views, Views data export, Features, Panels for hr16 config.
- Removing views-exposed-ajax-not-working-2425099-52.patch as it's not needed for current Views module version.


omarabuhussein:
looks good

PR #286, opened 2016-12-09
PCHR-1752 - Adding features_extra module and executing features-revert after enabling SSP features.

coldrunKacper:
I've added features_extra module to the drush.make.tmpl config to be able to export blocks using features (it requires to use fe_block submodule from features_extra).

Also I've added features reverting after SSP features are enabled in the install.sh script (to make sure that there are all feature code changes reflected in database).


colemanw:
@totten can you review this?

totten:
Looks reasonable/plausible. 

For anything coming from the CiviHR/Compucorp team and dealing with `app/config/hr*`, I tend to trust/accept their judgment. :)

PR #287, opened 2016-12-30
Adding no-sample-data option to install civi without sample data

omarabuhussein:
A new option is added (no-sample-date) which allows installing civicrm without any sample data (so installing civicrm with civicrm.sql only and without civicrm_generated.sql).

This option is also set to be disabled for hr16 build type by default.

Example usage :

typical installation **with** sample data (nothing changed)

```bash
civibuild create--type drupal-demo --url http://localhost:1234
```

OR 

```bash
civibuild create--type drupal-demo --url http://localhost:1234  --no-sample-data 0
```

Install **without** sample data :
```bash
civibuild create--type drupal-demo --url http://localhost:1234  --no-sample-data 1
```

### DIFF
--- a/src/civibuild.lib.sh
+++ b/src/civibuild.lib.sh
@@ -472,7 +472,9 @@ function civicrm_install() {
 
   pushd "$CIVI_CORE" >> /dev/null
     ## Does this build include development support (eg git or tarball-based)?
-    if [ -e "xml" -a -e "bin/setup.sh" ]; then
+    if [ "$NO_SAMPLE_DATA" == "1" ]; then
+      env SITE_ID="$SITE_ID" bash ./bin/setup.sh -Dgsdf
+    elif [ -e "xml" -a -e "bin/setup.sh" ]; then
       env SITE_ID="$SITE_ID" bash ./bin/setup.sh
     elif [ -e "sql/civicrm.mysql" -a -e "sql/civicrm_generated.mysql" ]; then
       cat sql/civicrm.mysql sql/civicrm_generated.mysql | eval mysql $CIVI_DB_ARGS
--- a/src/civibuild.lib.sh
+++ b/src/civibuild.lib.sh
@@ -472,12 +472,14 @@ function civicrm_install() {
 
   pushd "$CIVI_CORE" >> /dev/null
     ## Does this build include development support (eg git or tarball-based)?
-    if [ "$NO_SAMPLE_DATA" == "1" ]; then
+    if [ -e "xml" -a -e "bin/setup.sh" -n "$NO_SAMPLE_DATA" ]; then
       env SITE_ID="$SITE_ID" bash ./bin/setup.sh -Dgsdf
-    elif [ -e "xml" -a -e "bin/setup.sh" ]; then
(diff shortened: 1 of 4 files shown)


omarabuhussein:
Hello @totten  .. whenever you have time can you take a look at this ... can you please also merge this one (https://github.com/civicrm/civicrm-buildkit/pull/284/files) it is unlike this one , CiviHR only PR.

src/civibuild.lib.sh
```diff
@@ -472,7 +472,9 @@ function civicrm_install() {
 
   pushd "$CIVI_CORE" >> /dev/null
     ## Does this build include development support (eg git or tarball-based)?
-    if [ -e "xml" -a -e "bin/setup.sh" ]; then
+    if [ "$NO_SAMPLE_DATA" == "1" ]; then
```
totten: OK, makes sense that you'd call `setup.sh` with different args for this scenario.

With this patch, the `if`/`elif` chain is now addressing two distinct questions. These feel orthogonal from a user perspective, but each combination needs a slightly different implementation. The two questions are:

 * Are we working from raw source (i.e. using `git` to download `xml/schema/` and run `bin/setup.sh`) or from a published tarball (i.e. there's no `xml/schema/` but instead we have files in `sql/`)?
 * Do we want an empty/pristine DB or a populated/demo DB?

which leads to four use-cases:

 1. Have raw source (`-e xml -a -e bin/setup.sh`) and want sample data (`-z "$NO_SAMPLE_DATA"`)
 2. Have raw source (`-e xml -a -e bin/setup.sh`) and do not want sample data (`-n "$NO_SAMPLE_DATA"`)
 3. Have tarballs (`-e sql/civicrm.mysql`) files and want sample data (`-z "$NO_SAMPLE_DATA"`)
 4. Have tarballs (`-e sql/civicrm.mysql`) and do not want sample data (`-n "$NO_SAMPLE_DATA"`)

The current patch is going to confuse cases 2 and 4. One fix might be adding more conditionals and another branch in the `if`/`elif` chain.

Note: For case 4, the install logic should be `cat sql/civicrm.mysql sql/civicrm_data.mysql | eval mysql $CIVI_DB_ARGS`

src/civibuild.parse.sh
```diff
@@ -261,6 +261,11 @@ function civibuild_parse() {
         shift
         ;;
 
+      --no-sample-data)
```
totten: Whenever a new option is added, we should also describe the option in the help text, `civibuild_app_usage()`.

totten:
Thanks @omarabuhussein!

I ran into some issues well testing, so I've merged manually and included followups: e049f21e2617dbf44a1f37ddfaf3ab34b7a28602 and cddd92cd0a3ff993ab7389379aa3d2f21380cf54.

Notably, cddd92cd0a3ff993ab7389379aa3d2f21380cf54 changes/clarifies the signature slightly. It would affect your manual usage of the command, but I don't think it would affect configurations done through `src/civibuild.aliases.sh` or `app/civibuild.conf`.

omarabuhussein:
Thanks so much  @totten   , I was waiting to have sometime to fix this since I ran also into errors when I tested my last work but you beat me to it with some better tweaks :) .

PR #289, opened 2017-01-09
PCHR-1852: Removig unneeded drupal post-configurations file

omarabuhussein:
This script sets drupal prefixes to a specific list of hardcoded tables that are used in CiviHR self service portal , But this is not flexible and will cause problems when adding new custom groups since we will not be able to use them with drupal views module . So I removed the file from hr16 build type.


omarabuhussein:
@totten can you please merge this .. thanks

Issue #290, opened 2017-01-12
auto-branch bash script


xurizaemon:
From example in #155, a working implementation.

RM11632

xurizaemon:
Something bothers me about the trailing hyphens on these lines ... but it works OK. Just looks bothersome.

    do_auto_branch "backdrop" 1.x-
    do_auto_branch "drupal" 7.x-
    do_auto_branch "drupal" 6.x-

totten:
Added more feedback on #291.

PR #291, opened 2017-01-12
Script to facilitate easier branching.

xurizaemon:
Usage:

    cd build/dist/src
    civi-auto-branch origin/4.5.6 myorigin/4.5.6.7

Refs #290, #155.

### DIFF
--- a/bin/civi-auto-branch
+++ b/bin/civi-auto-branch
@@ -1,0 +1,50 @@
+#!/bin/bash
+
+function usage {
+  echo
+  echo Usage:
+  echo "  cd build/dist/src"
+  echo "  civi-auto-branch origin/4.5.6 myorigin/4.5.6.7"
+  echo 
+  echo "build/dist/src" contains the stock buildkit directories
+  echo "origin/4.5.6" refers to upstream branch origin
+  echo "myorigin/4.5.6.7" refers to your new branch
+  echo
+}
+
+if [[ ("$1" == "" || "$2" == "") ]] ; then
+  usage
+  exit 1
+fi
+
+if [[ ! ("$1" =~ .*/.* || "$2" =~ .*/.*) ]] ; then
+  usage
+  exit 1
+fi
+
+SRC_REMOTE=$(dirname $1)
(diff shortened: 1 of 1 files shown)


totten:
@xurizaemon  eah, I think this is a reasonable direction but want to talk through a couple things.

The script has a couple issues as written:
 * (Acceptable limitation) Only works in traditional D6/D7/distmaker/cividist structure -- it won't work in WP or D8 structure.
 * `civicrm-backdrop` is only applicable to v4.7. For v4.6 or earlier, it should be skipped. 
 * `civicrm-drupal` on v4.7 also needs the `8.x-` prefix.

FWIW, in `git scan` I tried to address this by having a "prefix detection" option in `git scan branch`, `git scan tag`, and `git scan push`. For example, see the instructions for [release candidates](https://wiki.civicrm.org/confluence/display/CRM/Release+Candidate). This scans your git repos/branches, prepares a plan of updates, and prompts you to accept/reject.

The `git scan` approach is more flexible, but it has its own issues:
 * (Acceptable limitation) There's a couple repos where we wind up making branches/tags that we don't really need (e.g. `zetacomponents/mail`).
 * It only works on local branches, and it's a bit futzy to get all the local branches up-to-date. I think that's solvable -- but probably more like "a couple hours work" rather than "5 minutes work". Maybe one of these:

I'm open-minded about either the `civi-auto-branch` approach or the `git scan` approach. IMHO options are:
 * Enhance `civi-auto-branch` to check the version number -- and to use slightly different steps in 4.6+4.7
 * Enhance `git scan up` so that one command updates all prefixes (e.g. `6.x-master` and `7.x-master` and `8.x-master`). (Basically, automate the "FIXME" the [wiki page](https://wiki.civicrm.org/confluence/display/CRM/Release+Candidate).)

totten:
For anyone who hasn't seen `git scan`:
 * README+Examples: https://github.com/totten/git-scan
 * Blog: https://civicrm.org/blogs/totten/developer-tip-managing-multiple-git-repositories

PR #295, opened 2017-01-21
Add in documentation on Druapl 8 build types in Civibuild now that mo…

seamuslee001:
…st parts of CiviCRM are working in Drupal 8

### DIFF
--- a/doc/civibuild.md
+++ b/doc/civibuild.md
@@ -60,8 +60,10 @@ configurations.  For example, at time of writing, it includes:
  * `backdrop-clean`: A bare, "out-of-the-box" installation of Backdrop+CiviCRM
  * `backdrop-demo`: A demo site running Backdrop and CiviCRM
  * `backdrop-empty`: An empty Backdrop site (without CiviCRM). Useful for testing tarball installation.
- * `drupal-clean`: A bare, "out-of-the-box" installation of Drupal+CiviCRM
- * `drupal-demo`: A demo site running Drupal and CiviCRM
+ * `drupal8-clean`: A bare, "Out of the box" Installation of Druapl8+CiviCRM.
+ * `druapl8-demo` : A demo site running Drupal8 and CiviCRM.
+ * `drupal-clean`: A bare, "out-of-the-box" installation of Drupal+CiviCRM.
+ * `drupal-demo`: A demo site running Drupal and CiviCRM.
  * `drupal-empty`: An empty Drupal site (without CiviCRM). Useful for testing tarball installation.
  * `joomla-empty`: An empty Joomla site (without CiviCRM). Useful for testing tarball installation.
  * `wp-demo`: A demo site running WordPress and CiviCRM


seamuslee001:
ping @totten @jackrabbithanna

jackrabbithanna:
Right now with the Drupal 8 installation process, the CiviCRM tables install in the Drupal db.  Is there a config option to make it install in a second, as we all use as standard?  If not, can the Civibuild stuff you are doing have a version that has it in 2 separate databases?  What's the estimated LOE to make that option, even if its a config option in settings.php or a yaml file somehere in the Drupal module

seamuslee001:
@jackrabbithanna the default with CiviBuild is it creates 3 databases even with the drupal8 builds 1 for the CMS 1 for CiviCRM 1 for a test DB to work with the unit tests

totten:
If you look at Civi across platforms, there's really a mix of different approaches to handling the single/split database issue:
 * **Drupal 6/7**: During installation, require the user to go to a special config screen. Always ask the user to put in two DSNs. (They can be the same DSN, but the UI doesn't suggest it.)
 * **WordPress**: During installation, require the user to go to a special config screen. Ask the user to put in two DSNs, but prepopulate both of them using WP's main DB.
 * **Joomla**: During installation, automatically install to Joomla's main DB.
 * **Drupal 8**: IIRC, during installation, it automatically installs to Drupal's main DB.
 * **civibuild**: During `civibuild install`, it creates three databases (`civi`, `cms`, `test`). The `civicrm_install` function initializes `civicrm.settings.php` to use them appropriately.

This is an uncomfortable mix for a couple reasons:
 * For typical D6/D7/WP admin, It requires extra installation steps and databases -- which feels unnatural to most Drupal/WordPress admins.
 * There are scenarios where admins prefer multiple DBs (mostly as a workflow/performance issue when you have different staff/consultants managing CMS and Civi). However, *cognition* of the trade-off usually comes *after you've used the system for a while*.
 * We wind up with very different installation experiences in each platform, which makes the docs more complicated.

The regime that I'd really like:
 1. **Default to simple**: In all CMS's, kill the special install screen. By default, install to the main CMS DB. (*However*, if `civicrm.settings.php` already exists, respect that. Also, don't try to re-install if any `civicrm_*` tables already exist.)
 2. **Power users**: Provide a command-line tool (probably in `cv`) to setup  `civicrm.settings.php` and the databases. This tool exposes more "power-user features" like splitting the databases, and it works on any CMS. This allows developers/sysadmins/devops folks to manage the installation in a more fine-grained way. 
 3. **Change whenever you want**: Provide a command-line tool (probably in `cv`)  to migrate `civicrm_*` tables/views/triggers between DBs. Whenever you realize that a different deployment is better, then you can switch to that.
 4. **Example data**: Losing the setup screen means that we don't get to ask about sample-data in the same way. Instead, the sample-data could be provided by an extension. (@michaelmcandrew has been working on a more powerful sample-data framework: https://github.com/civicrm/cv/pull/3). However, we'd probably still want some sign-posting after installation so that evaluators are aware of this.

totten:
@seamuslee001 thanks for docs update

PR #301, opened 2017-02-18
Update links to user guide

seamuslee001:
ping @totten does this link look right, Its either direct to the user guide or go to civicrm.org/documentation


totten:
This sounds good to me.

PR #307, opened 2017-03-23
Update wp-cli version to 1.1.0 - Better compatibility with WP 4.7

kcristiano:


totten:
Thanks @kcristiano . If `wp-cli` works for you, then it works for me. :)

Issue #315, opened 2017-05-02
outdated version of node/npm on Debian jessie leads to difficult mess


jmcclelland:
protractor has a dependency that uses the @ syntax, which npm version 1.4.21 can't parse, resulting in build failures (see debian bug:  https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=794890).

However, installing a new version of nodejs manually (per https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions) causes civi-download-tools to fail.

civi-download-tools fails because it detects jessie and tries to install npm and nodejs-legacy, however, the newer version of nodejs identifies these packages as "conflicts."

jmcclelland:
I think the proper fix is to run the nodejs setup script, then add nodejs as a package to install and remove npm and nodejs-legacy. 

Thoughts?

PalanteJon:
This is the approach taken when installing Ubuntu 12.04, and #318 is the same problem but for Ubuntu 14.04.  I could bikeshed the approach in #318, but it DOES work.  If @totten merges #318, I would take that as the go-ahead to submit a similar fix for Jessie.

jmcclelland:
Thanks Jon for the reference.

MikeyMJCO:
# Problem

**OS:** Debian Jessie 8.11.0
**Steps:**
 - Updated Jessie (`sudo apt update`) (`sudo apt upgrade`)
 - Installed Curl (`sudo apt install curl`)
 - Ran buildkit installer (`curl -Ls https://civicrm.org/get-buildkit.sh | bash -s -- --full --dir ~/buildkit`)

**Results:**
> protractor@4.0.14: wanted

> Error: Method Not Allowed

# Resolution

1. Remove nodejs-legacy and npm (`sudo apt remove nodejs-legacy npm`)
2. Install nodejs v8 from nodesource (`wget -O- https://deb.nodesource.com/setup_8.x | sudo -E bash -`)
3. Run `sudo apt install nodejs`
4. Re-run civi-download-tools (`cd ~/buildkit && ./civi-download-tools`)
5. Profit!

I'll raise a PR to get this fixed today.

MikeyMJCO:
PR is up #433

PR #318, opened 2017-05-03
Fix for ubuntu 14.04

steliosmilidonis:
The installation fails on ubutnu 14.04 , by installing nodejs 6 the installation runs very smooth

### DIFF
--- a/bin/civi-download-tools
+++ b/bin/civi-download-tools
@@ -321,7 +321,7 @@ function do_system_precise() {
 ###############################################################################
 function do_system_trusty() {
   set -e
-    PACKAGES="acl git wget unzip zip mysql-server mysql-client php5-cli php5-imap php5-ldap php5-curl php5-mysql php5-intl php5-gd php5-mcrypt php-apc apache2 libapache2-mod-php5 nodejs-legacy npm"
+    PACKAGES="acl curl git wget unzip zip mysql-server mysql-client php5-cli php5-imap php5-ldap php5-curl php5-mysql php5-intl php5-gd php5-mcrypt php-apc apache2 libapache2-mod-php5"
     echo "Detected \"Ubuntu Trusty 14.04\"."
     echo ""
     echo "Recommended packages: $PACKAGES"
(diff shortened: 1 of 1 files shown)


PalanteJon:
@totten While I could bikeshed this PR, it solves the problem and I'd recommend merging.  See also #315, which is this same problem on Debian Jessie.

totten:
I'm OK with this but would like comment from @AkA84 -- I think you said that the Shoreditch build depended on a particular version of `node`? Do you know which it was?

AkA84:
@totten yes it's v5.12.0 (see [here](https://github.com/civicrm/org.civicrm.shoreditch/blob/develop/CODING.md#build)). I haven't followed the issue, so i'm not sure if this is the most correct approach to the issue we can find. If it is, then I imagine that one could simply use [nvm](https://github.com/creationix/nvm) to *also* install v5.12.0 when he needs to work on Shoreditch

totten:
@steliosmilidonis @PalanteJon would node v5 work?

steliosmilidonis:
yes it would

bin/civi-download-tools
```diff
@@ -321,14 +321,16 @@ function do_system_precise() {
...
-    PACKAGES="acl git wget unzip zip mysql-server mysql-client php5-cli php5-imap php5-ldap php5-curl php5-mysql php5-intl php5-gd php5-mcrypt php-apc apache2 libapache2-mod-php5 nodejs-legacy npm"
+    PACKAGES="acl curl git wget unzip zip mysql-server mysql-client php5-cli php5-imap php5-ldap php5-curl php5-mysql php5-intl php5-gd php5-mcrypt php-apc apache2 libapache2-mod-php5"
     echo "Detected \"Ubuntu Trusty 14.04\"."
     echo ""
     echo "Recommended packages: $PACKAGES"
     echo ""
     if cvutil_confirm "Run automated installation? [Y/n] " y y; then
       sudo apt-get update
       sudo apt-get -y install $PACKAGES
+      sudo curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash -
```
PalanteJon: If you change `6.x` to `5.x` it sounds like this will get merged.

Hassan3152:
it works for me after doing the below steps:
1-run this script curl -Ls https://civicrm.org/get-buildkit.sh > get-buildkit.sh
2-search for (function do_system_trusty) and find (PACKAGES=) then remove nodejs-legacy and npm from this line and save it.
3-rub the rest of the step one script: bash civicrm-buildkit.sh --full --dir ~/buildkit
4-install nodejs 5 using this command: 
curl -sL https://deb.nodesource.com/setup_5.x | sudo -E bash - 
and the last step to setup nodejs 5 : sudo apt-get install -y nodejs

PR #324, opened 2017-05-11
closes #322 - don't assert vars in show --full

### LINKED ISSUE #322: "show --full" fails when any $PERSISTENT_VARS are not set

ginkgomzd:


totten:
Thanks. It's obviously smelly if `PERSISTENT_VARS` are missing, but it's not totally unheard-of (e.g. if we update list of `PERSISTENT_VARs`), and it does seem a bit much for `civibuild show` to crash.

PR #329, opened 2017-05-22
PCHR-2242: Add logintoboggan to modules to install

mickadoo:
## Overview
[LoginToboggan](https://www.drupal.org/project/logintoboggan) adds some features related to user login, for example _allowing users to login using either their username OR their e-mail address._

## Before
Users could only login using their username.

## After
Users can login using either username _or_ email

### DIFF
--- a/app/config/hr16/drush.make.tmpl
+++ b/app/config/hr16/drush.make.tmpl
@@ -175,6 +175,9 @@ projects[views_merge_rows][version] = "1.0-rc1"
 projects[masquerade][subdir] = civihr-contrib-required
 projects[masquerade][version] = "1.0-rc7"
 
+projects[logintoboggan][subdir] = civihr-contrib-required
+projects[logintoboggan][version] = "1.5"
+
 ; Patch for pagination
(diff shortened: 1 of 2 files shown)


mickadoo:
@totten can you or someone with merge rights please merge this?

eileenmcnaughton:
I think this is a case of both submitter & reviewer being from Compucorp is that correct? I'm happy to ignore normal rules about not having the reviewer from the same org as the submitter as this is a pretty straight forward change & I view Compucorp as the 'product owner' for CiviHR

eileenmcnaughton:
nb - there is no JIRA to close is there?

mickadoo:
@eileenmcnaughton thanks for merging this. I wasn't aware of the same org rule, but I'll keep it in mind in the future.

Re: Jira ticket, we just had an internal ticket (PCHR-2242) for this. If we need to I can create Jira tickets on https://issues.civicrm.org/jira/ in the future for similar features.

eileenmcnaughton:
In general the rules should be different orgs & always create a JIRA. There are definitely cases to make exceptions for on both rules!

PR #331, opened 2017-06-11
updated documentation again

jaapjansma:
On apache 2.4 the module apache_access_compat needs to be enabled.

### DIFF
--- a/doc/download-manual.md
+++ b/doc/download-manual.md
@@ -4,12 +4,12 @@ You may install buildkit in other environments. The main pre-requisites are:
 
  * Linux or OS X
  * Git
- * With apache version 2.4 you also need to enable mod_access_compat
- * PHP 5.3+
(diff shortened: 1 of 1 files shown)


totten:
Thanks. Reactions:
 * Yeah, makes sense to list required Apache modules.
 * The one that's screwed people the most has been `mod_rewrite`. (You can get through installation, but then Drupal's login form mysteriously fails.)
 * The statement "With apache version 2.4 you also need to enable mod_access_compat" was placed at a weird spot -- i.e. it was on the same level as "Linux", "git", "PHP", "NodeJS", "NPM". I moved it under the "Apache" requirement.
 * FWIW, I'm using Apache 2.4 on Ubuntu. `mod_access_compat ` is _not_ enabled (although it seems like the code is present). I'm curious why it was an issue - but don't mind including a tactful warning.

PR #338, opened 2017-07-13
PCHR-2387: Update Makefile to Install views_fieldsets v2.1

MiyaNoctem:
## Overview
Staging and demo servers were showing HTML code on /hr-resources view on attachment list for each resource.

image

## Before
Staging and demo servers were using the latest version of Visits_fieldsets module, which was incompatible with the template being used to render the HTML for the attachment list.

## After
Updated the original template so it's compatible with v2.1 of Views fieldset (see https://github.com/compucorp/civihr-employee-portal-theme/pull/188). Replaced v1.2 for v2.1 in drush makefile, so that now new installations use the latest version of the module.

image


totten:
The change is targeted at the CiviHR build, looks clean, and was approved by Davi. Merging.

PR #339, opened 2017-07-24
Remove warning about php-mysql during install

seancolsen:
I installed buildkit on a new system today and got hung up because of this warning message: 

> "Warning: This system will not support php-mysql extension"

I think this additional information is actually a disservice to the user because

* It doesn't explain how this affects the user. It's not actionable. 
* It causes additional stress during an already stressful process.
* It can be a red herring if the user has unrelated problems during the installation process (as was the case for me today)

I think if we're going to keep this warning message in, we should explain more to the user about the *implications* of the system not supporting php-mysql (which, personally, I don't understand).

### DIFF
--- a/bin/civi-download-tools
+++ b/bin/civi-download-tools
@@ -350,8 +350,6 @@ function do_system_xenial() {
     echo ""
     echo "Task: $PACKAGES"
     echo ""
-    echo "Warning: This system will not support php-mysql extension"
-    echo ""
     if cvutil_confirm "Run automated installation? [Y/n] " y y; then
       sudo apt-get update
       sudo apt-get -y install $PACKAGES


totten:
Let me give some explanation/context. Then we can see how it affects opinions:

 * IIRC, I added that message *before* [we decided on the eventual transition path for PHP 7](https://civicrm.org/blog/totten/psa-please-verify-php-extension-mysqli). At the time, it wasn't clear what that edge-case would mean.
 * Today, it's fair to say that the warning only matters if you're trying to install Civi v4.6.x or an early/historical release of 4.7.x.

@seanmadsen, it might be better to move or rephrase the message, but if you still prefer this, then I don't mind merging as-is (since the focus of work should be on `master`, where `php-mysql` doesn't matter).

seanmadsen:
@totten If we (you?) can re-phrase the message in a way the is more informative and actionable, then I would *prefer* to keep it in there. If you don't have ability/time to do that, then I think we should pull it out. I still don't understand enough about the issue to re-write the message myself. 

Related question: does this mean that I can't install CiviCRM 4.6 on my machine if I've seen this warning message during the install??

seamuslee001:
@seanmadsen at present 4.6 cannot be installed on any system that runs php7 only php5. There are open PRs to make 4.6 php7 compliant

MikeyMJCO:
Proposed rephrasing:

> Warning: This system will not support the php-mysql extension. This means you will be unable to install CiviCRM version 4.7.11 or earlier (including any 4.6.x version). For further details see: https://civicrm.org/blog/totten/psa-please-verify-php-extension-mysqli and https://issues.civicrm.org/jira/browse/CRM-17789

totten:
Yes, that's better language ^^^.

joemcl:
So #435 has been merged and this issue can be closed, @seancolsen ? Also - howdy :).

totten:
Previously substituted with #435.

PR #342, opened 2017-08-06
Link to Dev Guide for civicrm.settings.d docs

seanmadsen:
@totten [suggested](https://github.com/civicrm/civicrm-dev-docs/pull/217#issuecomment-314945998) this would be a good idea and I agree :slightly_smiling_face:


totten:
Thanks. Looks good. 👍

PR #343, opened 2017-08-08
PCHR-2485: Install yoti with drush on civihr installation

mickadoo:
## Overview
We are using the old version yoti-connect, checked into the civihr-employee-portal codebase. This is harder to maintain and will switch to using `drush` to install the module and also update the version.

## Before
The installation script did not install yoti, it was stored in civihr-employee-portal

## After
Yoti is installed and enabled as part of the civihr installation.

## Notes
It may be confusing that there are sometimes references yoti-connect and other times just 'yoti'. The module was renamed in a later version.


omarabuhussein:
looks good to me

totten:
Apologies - I didn't notice this when it first passed through. Merging.

PR #345, opened 2017-08-11
Remove phpunit3, replace with phpunit4 via composer

mattwire:
As discussed with @totten on chat.civicrm.org.  Remote phpunit "just works" with phpstorm when phpunit4 is installed via composer.  But I've been unable to get it working at all with phpunit3 via composer.
There doesn't seem to be any good reason to have two versions available.
It's a testing tool, so worst case a unit test breaks, that won't immediately affect the end-user so I think we can cope with that.

### DIFF
--- a/bin/civi-download-tools
+++ b/bin/civi-download-tools
@@ -22,7 +22,6 @@ CIVIXURL="https://download.civicrm.org/civix/civix.phar-2017-10-09-650c46da"
 CVURL="https://download.civicrm.org/cv/cv.phar-2017-10-02-2c279c8e"
 HUB_VERSION="2.2.9"
 DRUSH8URL=http://files.drush.org/drush.phar
-PHPUNITURL=https://phar.phpunit.de/phpunit-4.8.21.phar
 WPCLIURL=https://github.com/wp-cli/wp-cli/releases/download/v1.3.0/wp-cli-1.3.0.phar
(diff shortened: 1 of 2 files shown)


mattwire:
Need to run:
composer update phpunit/php-file-iterator phpunit/php-code-coverage phpunit/php-timer phpunit/php-token-stream phpunit/phpunit-mock-objects phpunit/dbunit phpunit/phpunit

seamuslee001:
@mattwire I don't believe this can happen until 4.6 goes EOL because 4.6 requires phpunit3, I believe @totten did much work as part of testapaloza that meant that 4.7 and up could run on phpunit4

totten:
There's a similar [discussion in #390](https://github.com/civicrm/civicrm-buildkit/pull/390#discussion_r186237206). The take-away is that we'll remove the `phpunit` copy provided by `composer.json`.

Regarding the last comment about v4.6 -- v4.6 has its own forked copy of phpunit 3. This copy (referenced by `composer.json`) has the distinction of being the first vanilla (unpatched) phpunit that worked with Civi, but it was only used for a brief period. (Once we got it working with vanilla, it became easier to try other versions).

Closing this PR. Will submit another one to remove `phpunit` from `composer.json`. At some point, we should also:

* Add a stub script (`bin/phpunit`; similar to `bin/codecept`) which delegates to a specific version.
* Change the file names for phpunit4/phpunit5 (and provide symlinks for compat) so that the `.phar` extension is visible. Helps with PHPUnit.

PR #346, opened 2017-09-03
Add in civi-test-run script

seamuslee001:
ping @totten

### DIFF
--- a/bin/civi-test-run
+++ b/bin/civi-test-run
@@ -1,57 +1,6 @@
 #!/bin/bash
 set -e
 
-function show_help() {
-  PROG=$(basename "$0")
-  echo "about: Execute the CiviCRM test suites"
-  echo "usage: $PROG <build-name> <junit-output-dir> <test-type>"
-  echo "test-type is one of all, karma, upgrade, phpunit-civi, phpunit-crm, phpunit-api, phpunit-e2e"
-}
-
-if [ -z $1 ]; then
-  echo "Missing argument: <build-name>"
-  echo
-  show_help
-  exit 2
-fi
-
-if [ -z $2 ]; then
-  echo "Missing argument: <junit-output-dir>"
-  echo
-  show_help
-  exit 2
-fi
-
-if [ -z $3 ]; then
-  echo "Missing argument: <test-type>"
-  echo
-  show_help
-  exit 2
-fi
-
-## Determine the absolute path of the directory with the file
-## absdirname <file-path>
-function absdirname() {
-  pushd $(dirname $0) >> /dev/null
-    pwd
-  popd >> /dev/null
-}
-
-BINDIR=$(absdirname "$0")
-PRJDIR=$(dirname "$BINDIR")
-BLDNAME="$1"
-BLDDIR="$PRJDIR/build/$BLDNAME"
-CIVI_CORE="$BLDDIR/sites/all/modules/civicrm"
-JUNITDIR="$2"
-TESTTYPE="$3"
-EXITCODES=
-GUARD=
-
-#################################################
-## Cleanup left-overs from previous test-runs
-[ -d "$JUNITDIR" ] && $GUARD rm -rf "$JUNITDIR"
-[ ! -d "$JUNITDIR" ] && $GUARD mkdir "$JUNITDIR"
-
 #################################################
 ## Helpers
 function fatal() {
@@ -71,8 +20,18 @@ function getCiviVer() {
   popd >> /dev/null
 }
 
+function show_help() {
+  PROG=$(basename "$0")
+  echo "about: Execute the CiviCRM test suites"
+  echo "usage: $PROG <build-name> <junit-output-dir> <test-type>"
+  echo "test-type is one of all, karma, upgrade, phpunit-civi, phpunit-crm, phpunit-api, phpunit-e2e"
+}
+
 #################################################
 ## Test tasks
+##
+## Each task is responsible for depositing a JUnit XML file under
+## "$JUNITDIR".
 
 function task_phpunit() {
   case "$TESTTYPE" in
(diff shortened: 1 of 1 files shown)


bin/civi-test-run
```diff
@@ -0,0 +1,116 @@
+#!/bin/bash
+set -e
+
+if [ -z $1 ]; then
+  echo "No Buildkit path passed in"
```
totten: This part can be determined automatically. Copy/paste the preamble from one of the other bash scripts, eg

https://github.com/civicrm/civicrm-buildkit/blob/1a9922da3bec95b7b9d5bb9731f31ce193a8b756/bin/civi-download-tools#L8-L17

(`PRJDIR` in `civi-download-tools` is the same as `BLDKIT` in `civi-test-run`.)

bin/civi-test-run
```diff
@@ -0,0 +1,116 @@
...
+  *)      echo "UpgradeTest failed: Unrecognized version ($CIVIVER)" ; exit 1; ;;
+  esac
+if  ! $GUARD civibuild upgrade-test $BLDNAME $UTVERS ; then
+    EXITCODES="$EXITCODES upgrade-test"
+fi
+  $GUARD cp "$BLDKIT/app/debug/$BLDNAME/civicrm-upgrade-test.xml" "$JUNITDIR/"
+  $GUARD civibuild restore "$BLDNAME"
+fi
+
+runPHPUnit $TESTTYPE $EXITCODES $GUARD
```
totten: By default, bash variables are global -- there should be no need to pass `EXITCODES` or `GUARD`.

bin/civi-test-run
```diff
@@ -0,0 +1,116 @@
...
+
+#################################################
+## Helpers
+function fatal() {
+  echo "$@"
+  exit 1
+}
+
+## RUN PHP Unit tests
+function runPHPUnit() {
```
totten: The use of `$1`, etal, is a bit obscure -- adding some sort of docblock/signature would help. e.g.

```bash
## Execute a PHPUnit suite
## 
## Usage: runPHPUnit <test>
## Ex: runPHPUnit PHPUNIT-API
function runPHPUnit() { 
```

bin/civi-test-run
```diff
@@ -0,0 +1,116 @@
...
+  case "$1" in
+  PHPUNIT-API*) TESTS="api_v3_AllTests" EXITCODE="$EXITCODES phpunit API" ;;
```
totten: `<bike-shedding>` For a naming convention, consider that this is something you might type into a console. Go lowercase? `</bike-shedding>`

bin/civi-test-run
```diff
@@ -0,0 +1,116 @@
...
+set -e
+
+if [ -z $1 ]; then
+  echo "No Buildkit path passed in"
+  exit 2
+fi
+
+if [ -z $2 ]; then 
+  echo "No Build Name passed in"
+  exit 2
```
totten: Before exiting, call a function to show the full help screen, ex

```bash
function show_help() {
  echo "Execute the CiviCRM test suites"
  echo "usage: $0 <build-name> <civi-ver> <junit-output-dir> <test-type>"
}
```

bin/civi-test-run
seamuslee001: @totten this is no longer needed i believe

bin/civi-test-run
```diff
@@ -0,0 +1,133 @@
...
+    EXITCODES=$EXITCODE
+  fi
+  phpunit-xml-cleanup "$JUNITDIR"/*.xml
+}
+
+## Run Karma tests if requested Javascript
+if [ $TESTTYPE = "karma" ] || [ $TESTTYPE = "all" ] ; then
+  $GUARD pushd "$BLDDIR/sites/all/modules/civicrm"
+    ## Work-around: ensure pradmin user has sa contact
+    drush -u pradmin cvapi contact.get id=user_contact_id
```
seamuslee001: @totten this account may not always be set, so not sure what to do here

seamuslee001:
@totten i think this script is pretty good now, the only issue is around L81

bin/civi-test-run
totten: @seamuslee001 If we moved `civi-test-run` into a subcommand of `civibuild` (eg `civibuild test dmaster ...` -- stored in `src/command/test-run.sh`), then `$ADMIN_USER` would be available implicitly, but that might be a bit more work to make it a subcommand.

This might work as a drop-in change:

```
ADMIN_USER=$(cv ev 'echo $GLOBALS['_CV']['ADMIN_USER'];')
drush -u "$ADMIN_USER" cvapi contact.get id=user_contact_id
```

or maybe this:

```
cv ev 'CRM_Utils_System::synchronizeUsers();'
```

seamuslee001:
@totten i think i got the Karma tests running locally, Question will be is if this can work for Jenkins

totten:
@seamuslee001 I've pushed a few updates (some fixes and some tweaks -- 6d0cc4e..e51f504). I've spot-checked these commands, and they seem to be working on my local:
 * `civi-test-run dcase /tmp/out phpunit-civi`
 * `civi-test-run dcase /tmp/out karma`
 * `civi-test-run dcase /tmp/out upgrade`

If the changes are OK, then I'm inclined to merge so that we can setup an experimental Jenkins job and see how well it runs with the changes.

Note: although I'm +1 for merging, I'd still like to treat it as experimental and hopefully make a subsequent change to the signature so that we might say:
 * Signature: `civi-test-run -b BLDNAME -j JUNITDIR <TEST1> <TEST2...>`
 * Nightly tests run with `civi-test-run -b dmaster -j /tmp/output karma upgrade-all phpunit-{e2e,civi,crm,api}`
 * PR tests run with: `civi-test-run -b dmaster -j /tmp/output karma upgrade-lite phpunit-{e2e,civi,crm,api}`

Differences: 
 * You can give a list with multiple suites to run. To make the arguments less ambiguous, the build name and junit dir become optional. When running in alternate configurations (like one might do for the CMS-integration projects), you can specify whatever mix of suites you want.
    * (I appreciate that the command can be called separately for each suite, but this formulation ensures that the exit-reporting is done consistently across all the suites.)
 * Full/scheduled/matrix jobs run the upgrade test with all `4.x.y` snapshots, and PR jobs only run with the `4.x.*` snapshots. (This is actually a difference in how `CiviCRM-Core-Matrix` and `CiviCRM-Core-PR` run today.)
 * In the future, we could probably infer `-b BLDNAME` from the CWD (much like `civibuild` can).

seamuslee001:
@totten i have looked over the changes and they seem to make sense. Agree with your comments. I think something like the packages repo may be a nice one to experiment with. That would allow us to keep PR and normal matrix running as per normal

seamuslee001:
Doing a test run locally on the changes it all seems to be working well running as all it proceeded doing Karma then Upgrade then E2E then CRM and all seems well without causing error so i think its good to merge noting the experimental nature of this thing

PR #347, opened 2017-09-07
WIP General updates - UK Sprint 2017?

mattwire:
Changes for discussion, not merge ready yet.  Possibly worth looking at for UK Sprint 2017 - @totten?
* Install drush, wp-cli via composer.
* Install phpunit4 via composer: #345
* Use nodejs v5: See #318 and https://github.com/civicrm/civicrm-buildkit-vagrant/pull/23
* Add support for debian stretch (PHP7.0 + mariadb)

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,3 +40,14 @@ see [CiviCRM Developer Guide: Buildkit](https://docs.civicrm.org/dev/en/latest/t
     * [`karma`](http://karma-runner.github.io) (w/[jasmine](http://jasmine.github.io/)) - Unit testing for Javascript.
     * [`paratest`](https://github.com/brianium/paratest) - Parallelized version of PHPUnit.
     * [`phpunit` and `phpunit4`](http://phpunit.de/) - Unit testing for PHP (with Selenium and DB add-ons).
+    
+## Stretch Install
+By default you can login to mysql with no password as root.  To change this:
+```
+sudo mysqld_safe --skip-grant-tables &
+mysql -u root
+use mysql;
+update user set password=PASSWORD("mynewpassword") where User='root';
+update user set plugin="mysql_native_password";
+quit;
+```
--- a/bin/civi-download-tools
+++ b/bin/civi-download-tools
@@ -501,7 +501,7 @@ function do_system_jessie() {
 function do_system_stretch() {
   set -e
     PACKAGES="acl git wget unzip zip mysql-server mysql-client php-cli php-imap php-ldap php-curl php-mysql php-intl php-gd php-mcrypt php-apcu apache2 libapache2-mod-php php-bcmath php-mbstring php-soap php-zip"
-    echo "Detected \"Debian Jessie\"."
+    echo "Detected \"Debian Stretch\"."
     echo ""
     echo "Installing nodejs Debian repository."
     wget -O- https://deb.nodesource.com/setup_8.x | sudo -E bash -


kcristiano:
@mattwire Thanks for the work here!   One comment is that we have had issues pulling wp-cli via composer before due to conflicts with dependencies.   The conflict was with the joomla cli utility.  We can test and see if this is resolved, but is there a benefit to switching away from the phar download?

mattwire:
@kcristiano Thanks for looking at it :-)  There's no real reason to move away from the phar download other than trying to move more of the scripted stuff to composer instead.  Useful to know there was an issue and would be great if you could test to see if the issue still exists.

kcristiano:
@mattwire I'll get a machine running where I can test.  I just set my laptop up for the MD Sprint so I want to leave that as is.  I had planned updating the wp-cli phar version to 1.3.0, so let's see if we can move wp-cli to composer.

bin/civi-download-tools
```diff
@@ -21,9 +21,6 @@ LOCKTIMEOUT=90
 CIVIXURL="https://download.civicrm.org/civix/civix.phar-2017-08-02-2489efea"
 CVURL="https://download.civicrm.org/cv/cv.phar-2017-07-27-89ba2508"
 HUB_VERSION="2.2.9"
-DRUSH8URL=http://files.drush.org/drush.phar
-PHPUNITURL=https://phar.phpunit.de/phpunit-4.8.21.phar
-WPCLIURL=https://github.com/wp-cli/wp-cli/releases/download/v1.1.0/wp-cli-1.1.0.phar
```
totten: As @kcristiano said:

> One comment is that we have had issues pulling wp-cli via composer before due to conflicts with dependencies.

Originally, `buildkit` downloads were almost entirely handled via `composer`, but that style of distribution basically requires that *all CLI tools* use the same versions of their libraries. The `composer` global style has the same problem. PHAR-based distribution solves this. (I have a fairly strong suspicion that this is why most PHP-CLI tools recommend their PHAR-based builds.)

It would be prettier if these requirements were listed in some kind of JSON file. You could potentially write a small PHP script and JSON file which handles this in a nicer fashion. If you really wanted it in`composer.json`, it be a [plugin](https://getcomposer.org/doc/articles/plugins.md). Ex:

```json
  "require": {
    "composer/phar-plugin" : "^1.0"
  },
  "extra": {
    "phar": {
      "bin/phpunit4": "https://phar.phpunit.de/phpunit-4.8.21.phar",
      "bin/wp": "https://github.com/wp-cli/wp-cli/releases/download/v1.1.0/wp-cli-1.1.0.phar"
    }
  }
```

(Update: These look interesting but seem to require PHP 5.6: https://github.com/tommy-muehle/composer-tool-installer-plugin and https://github.com/tommy-muehle/tooly-composer-script )

bin/civi-download-tools
```diff
@@ -164,18 +161,6 @@ function show_command() {
 }
 
 ###############################################################################
-## Debian.org's NodeJS package uses a non-standard name for the node binary.
-## If necessary, setup an alias for the standard name.
-function nodejs_debian_workaround() {
```
totten: In the past, we had repeated support issues from Debian/Ubuntu users. It's not a matter of design-aesthetics (by which this function is ugly and properly redflagged as `_workaround`) -- it's a matter of pain-avoidance.

If you have the perspective that it's normal pull in a pre-specified stack (e.g. `--full` or `docker` or `vagrant`), then this wouldn't be necessary. (You could choose full-stacks that provide a normal `node` binary.) But IMHO it's quite useful to be able to run this stuff on different stacks, and sometimes those stacks will be based on Debian/Ubuntu, and often those systems will screw-up the name of the `node` binary.

README.md
```diff
@@ -40,3 +40,14 @@ see [CiviCRM Developer Guide: Buildkit](https://docs.civicrm.org/dev/en/latest/t
...
+    
+## Stretch Install
+By default you can login to mysql with no password as root.  To change this:
+```
+sudo mysqld_safe --skip-grant-tables &
+mysql -u root
+use mysql;
+update user set password=PASSWORD("mynewpassword") where User='root';
+update user set plugin="mysql_native_password";
+quit;
```
totten: 1. This is curious information. What's the `mysql_native_password` stuff about?
 2. The installation docs for buildkit were moved to https://github.com/civicrm/civicrm-dev-docs/blob/master/docs/tools/buildkit.md -- so it should be addressed over there .

PR #349, opened 2017-09-12
civi-test-run - Accept a list of multiple suites

seamuslee001:
@totten local testing suggests this is doing sensible things. in that if i tell it to run something like 

```civi-test-run -b 47.demo -j /home/seamus/junit karma phpunit-e2e phpunit-civi``` it only runs karma phpunit-e2e phpunit-civi and it does appear to clean out the junit dir as well

Let me know if i need to do anything more but i think this will make it more suitable for Jenkins

### DIFF
--- a/bin/civi-test-run
+++ b/bin/civi-test-run
@@ -93,28 +93,11 @@ function task_upgrade() {
 }
 
 #################################################
-## Parse inputs
-
-if [ -z $1 ]; then
-  echo "Missing argument: <build-name>"
-  echo
-  show_help
-  exit 2
-fi
-
-if [ -z $2 ]; then
-  echo "Missing argument: <junit-output-dir>"
-  echo
-  show_help
-  exit 2
-fi
-
-if [ -z $3 ]; then
-  echo "Missing argument: <test-type>"
-  echo
-  show_help
-  exit 2
-fi
+## Cleanup left-overs from previous test-runs
+function junit_cleanup() {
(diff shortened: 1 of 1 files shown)


totten:
Looks quite improved. 👍 

I made a couple more changes which are described in the git commits above.

Stylistically, it feels like the script could be simplified -- e.g.
 * Instead of checking for `TESTTYPE = all` in many places, the parse should check `$1 = all` and then set `SUITES="karma upgrade phpunit-e2e..."`. Everything else should accept multiple items from `SUITES`, so they don't need to explicitly check for `all`.
 * Perhaps combine the `case` from `task_phpunit` with the `if` from `run_tests` and produce one dispatch function along the lines of:
   ```bash
   for TESTTYPE in $SUITES ; do
     case "$TESTTYPE" in
       karma)  task_karma ;;
       upgrade)  task_upgrade ;;
       phpunit-e2e)  task_phpunit E2E_AllTests ;;
       phpunit-crm)  task_phpunit CRM_AllTests ;;
       phpunit-api)  task_phpunit api_v3_AllTests ;;
       phpunit-civi)  task_phpunit 'Civi\AllTests' ;;
       *) echo "unrecognized suite: $TESTTYPE"; exit 2 ;;
     esac
   done
   ``` 

However, that's not a blocker. If you're OK with the other updates above, then we can merge.

seamuslee001:
I think the changes are fine and i tested them locally, agree now we have suites that might be best but maybe another iteration, This should at least allow us to expand what we can run for drupal and start setting up packages right?

totten:
Yupyup.

PR #351, opened 2017-09-13
civi-test-run - Simplify dispatch of subtasks

seamuslee001:
… and consolodation of calling functions

### DIFF
--- a/bin/civi-test-run
+++ b/bin/civi-test-run
@@ -50,28 +50,16 @@ function show_help() {
 ## "$JUNITDIR".
 
 function task_phpunit() {
-  case "$TESTTYPE" in
-    phpunit-api*) TESTS="api_v3_AllTests" EXITCODE="$EXITCODES phpunit API" ;;
-    phpunit-civi*) TESTS="Civi\\AllTests" EXITCODE="$EXITCODES phpunit Civi" ;;
-    phpunit-crm*) TESTS="CRM_AllTests" EXITCODE="$EXITCODES phpunit CRM" ;;
-    phpunit-e2e*) TESTS="E2E_AllTests" EXITCODE="$EXITCODES phpunit E2E" ;;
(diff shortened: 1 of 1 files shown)


bin/civi-test-run
```diff
@@ -175,6 +144,14 @@ while [ -n "$1" ] ; do
     show_help
     exit 2
 
+  elif [ $1 = "all" ] ; then
+    if [ -f "$CIVI_CORE/tests/phpunit/E2E/AllTests.php" ]; then
```
totten: I like how this makes "all" do the Right Thing in more cases.

I'm less enthused about how it makes the order of arguments subtly more brittle -- eg these would behave differently:
 * `civi-test-run -b foo all` (includes or excludes E2E depending on the content of `foo`)
 * `civi-test-run all -b foo` (never runs E2E, regardless of version)

seamuslee001: hmmm

totten: Well, if you need per-build detection, then it can be fixed after the loop. It's a little less pretty but should be fine:

```bash
SUITES=
USE_ALL=
while [ -n "$1" ] ; do
  #...parsing...
  if [ "$1" = "all" ]; then
     USE_ALL=1
  fi
  #...parsing...
done

if [ "$USE_ALL"]; then
  # ...compute $SUITES...
fi
```

totten:
Definitely easier to trace this updated code. :)

totten:
Seems good to merge.

Some test results:
 * `civi-test-run -b dmaster -j /tmp/junit phpunit-e2e` (works)
 * `civi-test-run -b wpmaster -j /tmp/junit phpunit-e2e` (sorta works, sort doesn't -- but i think that's an actual issue with the code)
 * `civi-test-run -b bmaster -j /tmp/junit phpunit-e2e` (sorta works, sorta doesn't -- but i think that's an actual issue with the code)
 * `civi-test-run -j /tmp/junit46 all -b d46` (seems to work; picks the right suites; but i didn't let it execute to completion)
 * `civi-test-run -j /tmp/junit46 all -b dmaster` ((seems to work; picks the right suites; but i didn't let it execute to completion)

PR #353, opened 2017-09-15
Civi-test-run add in function to run phpunit tests within the drupal …

seamuslee001:
…folder

ping @totten the only issue for me is that should i use phpunit4 or phpunit for this. The later is probably consistent with 4.6 but not 100% sure

### DIFF
--- a/bin/civi-test-run
+++ b/bin/civi-test-run
@@ -104,7 +104,7 @@ function junit_cleanup() {
   [ ! -d "$JUNITDIR" ] && $GUARD mkdir "$JUNITDIR"
 }
 
-function phpunit_drupal() {
+function task_phpunit_drupal() {
   $GUARD pushd "$CIVI_CORE/drupal"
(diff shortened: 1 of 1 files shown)


bin/civi-test-run
```diff
@@ -179,6 +189,7 @@ for TESTTYPE in $SUITES ; do
     phpunit-crm)  task_phpunit CRM_AllTests ;;
     phpunit-api)  task_phpunit api_v3_AllTests ;;
     phpunit-civi)  task_phpunit 'Civi\AllTests' ;;
+    phpunit-drupal) phpunit_drupal ;;
```
totten: @seamuslee001 Nice! I'd really want to get the `civicrm-drupal` suite running.

Minor nitpick: the other functions have a naming convnetion (`task_karma`, `task_upgrade`, `task_phpunit`). This one should be `task_phpunit_drupal`.

seamuslee001: changed in the latest commit @totten

seamuslee001:
@totten I think this is good to merge now Tim

totten:
Cool :)

PR #354, opened 2017-09-18
WMFF config: add extensions directory to civicrm.settings.php

ejegg:
Pretty ugly, but it does the trick in our CI when other attempts fail.

There's got to be a better way to do this, right?


ejegg:
@eileenmcnaughton unfortunately the setting creation api call to handle this seems not to do the trick. Looks like our CI tests had been relying on this modification to files on the CI servers, and when they cleared out the workspace all the CRM tests started failing.

ejegg:
Oops, found the right way to do this

PR #355, opened 2017-09-18
WMFF config: Add CiviCRM extensions directory and URL

ejegg:


eileenmcnaughton:
Nice - merging as per previous discussions with @totten this folder of buildkit is a wmf specific build & we get to do what we want there :-)

PR #358, opened 2017-09-28
Update wp-cli to v1.3, fix softlink errors, specify default WP Version

kcristiano:


app/config/wp-demo/download.sh
```diff
@@ -5,15 +5,14 @@
 ###############################################################################
 [ -z "$VOL_VERSION" ] && VOL_VERSION='4.4-1.x'
 
-WPCLI_ARGS=
-[ -n "$CMS_VERSION" ] && WPCLI_ARGS="$WPCLI_ARGS --version=$CMS_VERSION"
+[ -z "$CMS_VERSION" ] && CMS_VERSION=4.8
```
totten: So just thinking out loud -- the basic trade-off here is:
 * If the default is to build with the latest version of WordPress, then we'll find out sooner if things are broken. But things might be broken from time-to-time!
 * If the default is to build with a specific version of WordPress, then it works more reliably. But we may be blissfully ignorant about new compatibility problems!

I can go along with this change... but want to make sure we have a likely path to discovering when new problems arise. Perhaps a more balanced regime:

 * `civibuild` would be pegged to a version (eg 4.8, as above)
 * The CI job -- https://test.civicrm.org/job/CiviCRM-Core-Matrix/ -- would have a dimension for testing different CMSs (ie `civibuild create wp-demo && civi-test-run ...` and `civibuild create backdrop-clean && civi-test-run ...`). The CI job could be configured to follow the latest version.

To allow for that regime... @kcristiano, would you expect that `civibuild create wpmaster --cms-ver latest` would have the intended effect?

kcristiano: Yes I would expect latest to pull the latest.   I want to pin to the major version due to upcoming BC on the roadmap for WP.   While I want to build with these changes for testing, I do not see the need to break our buildkit before it's time.   Added to that the 4.8.2 wp-cli download is marked as corrupt on Ubuntu so new wpmaster builds are failing now.  So I figured the time was ripe to explicitly set a version and then let folks that want the latest get that with the --cms-ver flag.

kcristiano: also FYI --cms-ver nightly still lets us get the pre-release alpha/beta.

totten: Okie dokie. Merged!

Issue #364, opened 2017-10-22
Silence which nodejs error message


michaelmcandrew:
https://github.com/civicrm/civicrm-buildkit/commit/09682c7c92b5f8a3b050c045956fb90631ffaa64 introduces a workaround for the node/nodejs naming issue.

An side effect of this is the printing of an error message along the lines of 

```
which: no nodejs in (/usr/local/sbin:/usr/local/bin:/usr/bin...
```

when nodejs is not found. This error message can be silenced by redirecting the error (along with the output) to /dev/null.

A minor improvement but increases our signal to noise ratio which is good for usability.

totten:
Merged the PR. Thanks!

PR #365, opened 2017-10-22
redirect which nodejs error message to /dev/null

michaelmcandrew:
see issue #364

### DIFF
--- a/bin/civi-download-tools
+++ b/bin/civi-download-tools
@@ -167,8 +167,8 @@ function show_command() {
 ## Debian.org's NodeJS package uses a non-standard name for the node binary.
 ## If necessary, setup an alias for the standard name.
 function nodejs_debian_workaround() {
-  if which nodejs >> /dev/null ; then
-    if ! which node >> /dev/null ; then
+  if which nodejs &> /dev/null ; then
+    if ! which node &> /dev/null ; then
       echo "[[NodeJS binary appears to be misnamed. Creating 'node' alias.]]"
       ln -s "$(which nodejs)" "$BINDIR/node"
     fi


totten:
Oh, neat -- I hadn't seen the `&>` operator before. I've checked that notation in a few different versons of `bash`, and it seems to work as expected. 👍

michaelmcandrew:
cool :)

ginkgomzd:
It might be old-skool to be careful about doing output redirection with '&>'... but historically, it was not regarded as good practice for portability.
https://stackoverflow.com/questions/6674327/redirect-all-output-to-file#comment26613349_6674383

michaelmcandrew:
Hey @GinkgoFJG,
Since that script is specified as bash, I think we should be OK using bashisms.

https://github.com/civicrm/civicrm-buildkit/blob/4632601652874ea4c8b5111aaf188301e82fe505/bin/civi-download-tools#L1

GinkgoFJG:
I think the previous comment was meant for @ginkgomzd.

michaelmcandrew:
indeed it was, sorry @GinkgoFJG :)

ginkgomzd:
It's not about Bash vs another shell.

@totten, you said you you tested on a few... did that include windows/cygwin?
That is the only platform I would have concerns about being a weenie-head POSIX-only bash implementation.

Again, might be old info... I forgot that the Windows Subsystem for Linux exists now...

hmmm... M$ support article does not include it as an option for redirection:
https://support.microsoft.com/en-us/help/110930/redirecting-error-messages-from-command-prompt-stderr-stdout

P.S. - a better reference for this being risky syntax:
http://wiki.bash-hackers.org/scripting/obsolete

totten:
The `bash-hackers.org` article is pretty cool in highlighting non-portable syntax (although, to my reading, their pitch is grounded in portability concerns). It would be useful to follow those recommendations in a few circumstances like:

 * Documentation for admins/users (where the readers might be using different shells)
    * Oh, speaking of which, let's do this: https://github.com/civicrm/civicrm-dev-docs/pull/432
 * Autogenerated shell code (where a command like `amp export -Ncms ` or `docker-machine env default` prints shell code that should be evaluated by the user's preferred shell)
 * Example sysinit scripts bundled with a daemon (where downstream distros might try to micro-optimize boot-times by replacing `bash`)

But this case is not any of those -- `civibuild` is basically an application written in the `bash` language. When writing a non-trivial application as a shell script, anyone who values their sanity should pick a particular dialect.  As @michaelmcandrew points out, `bash`-isms are safe in `bash`. (Note the first line of `bin/civibuild` explicitly binds to bash: `#!/bin/bash`. If someone jerry-rigs `/bin/bash`, then they deserve whatever pain comes to them.)

The MS article references `dir`... so that's probably `cmd.exe` or PowerShell, which are entirely different beasts. Cygwin and WSL provide `bash`. (There may be problems with running `civibuild` in Cygwin or WSL... but not because they mix-up the shell interpreters.)

ginkgomzd:
ok, so maybe it _is_ about bash vs something else. It all seems like a solid case. I have to reserve a modicum of scepticism just because I've long used she-bangs and still run into unexpected problems when moving to a different server... so forgive me for dogmatically heeding "portability warnings".
I've literally wanted to use this for years... so maybe I will start being braver about it. 

Okay... thinking about times I know I've been burned... I just logged into a production server and found that indeed... the bash version is 3.x. The language feature in question was not supported until 4.x.
If one major hosting provider thinks bash 3.x is "good enough" in 2017, I'd warrant there are others.
This is a production server though, and I'll admit that any development environment is going to have a better chance of having up-to-date software.

Anyway... this was fun and educational.
Cheers

Issue #369, opened 2017-10-25
civicrm/packages overwritten during civibuild create


michaelmcandrew:
I am having a fairly hard time understanding how this drush.make.tmpl has been working for people (myself included! - it was working fine, then stopped)when libraries[civicrm] is defined after libraries[civicrmpackages] etc. When I run it, the civicrm/packages folder (and similar folders like civicrm/l10n, civicrm/drupal etc.) are overwritten when the libraries[civicrm] definition is processed. As a result, I get a PHP Warning:  require_once(Log.php): failed to open stream: No such file or directory in /var/civicrm-buildkit/build/dmaster/sites/all/modules/civicrm/CRM/Core/Config.php on line 38 error (the file can normally be found in civicrm/packages) and civibuild create dies half way through... 

Re-arranging the order of definitions in the file so that libraries[civicrm] appears before libraries[civicrmpackages] fixes the problem for me.

I tested this with drush 7.0-dev and 8.1.15.

michaelmcandrew:
I'm not 100% on the testing on both versions - it *might* be something that only happened for me in 81.15 but at the minute I am fairly sure it happen in both versions. I can confirm that if you like.

totten:
I've only seen the problem on `drush8` -- with `drush` 7.0-dev, it's always worked for me.

Never-the-less, reordering it is a good idea. (I think that v7 works either way.)

totten:
I think this was resolved by #370.

PR #370, opened 2017-10-26
Sort drush.make "libraries[]"

michaelmcandrew:
I have re-ordered the 'libraries' in the makefiles as outlined in https://github.com/civicrm/civicrm-buildkit/issues/369 with the exception of D8 since I am not familiar with these and there seemed to be some inconsistencies in the make files.

Compare

https://github.com/civicrm/civicrm-buildkit/blob/4632601652874ea4c8b5111aaf188301e82fe505/app/config/drupal8-clean/drush.make.tmpl#L27

with 

https://github.com/civicrm/civicrm-buildkit/blob/4632601652874ea4c8b5111aaf188301e82fe505/app/config/drupal8-demo/drush.make.tmpl#L27

I wasn't sure which of these was correct. Maybe both? maybe neither! I will do a little bit more testing...

### DIFF
--- a/app/config/doctrine/drush.make.tmpl
+++ b/app/config/doctrine/drush.make.tmpl
@@ -22,6 +22,13 @@ projects[] = drupal
 ; This will make it easier to submit pull-requests for patches.
 ; see: http://wiki.civicrm.org/confluence/display/CRMDOC/Github+for+CiviCRM
 
+libraries[civicrm][destination] = modules
+libraries[civicrm][directory_name] = civicrm
+libraries[civicrm][download][type] = git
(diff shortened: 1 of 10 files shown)


michaelmcandrew:
PS. while I was there, I also ensured that `civicrm/l10n/fr_CA/LC_MESSAGES` came after `civicrm/l10n`. I appreciate that these are commented out and might not be in use any more but still it is good if the files we distribute work properly (if and when we switch to drush8).

michaelmcandrew:
My experiments with building Drupal 8 were not successful. I am guessing that  I am either missing some dependencies or those build configs are not working at the moment.

totten:
For the build-types that are included in the PR, the changes make sense. I spot-checked with `drupal-demo` and that worked fine. :+1:

Issue #368, opened 2017-10-23
Consider not altering $PATH in bin/civibuild


michaelmcandrew:
The following lines are a belt and braces approach to ensuring that civibuild has access to the binaries that it needs:

https://github.com/civicrm/civicrm-buildkit/blob/823b193cb5618fbe42284fc9e2dfbb9e58a3ae02/bin/civibuild#L20-L21

But they prevent people who 'know better' from allowing their own versions of civibuild's tools. For example, I would like to run a forked version of amp with my own nginx config template but am unable to do so since the path is overwritten.

https://docs.civicrm.org/dev/en/latest/tools/buildkit/#path already mentions that you need to configure your $PATH correctly which is enough and less intrusive.

In the comment you wrote 'available'. If you wanted to ensure that these specific versions were used then this is a good approach. If you just want them as a backup, you could always put them at the end. Or if we decided that rtf-installation-m was the best way to go then I would advocate removing it entirely.

totten:
Yeah, when I first added that, setting the `PATH` like this killed two birds with one stone:

 * It allows `civibuild` to work even if you don't put `buildkit/bin` in the `PATH`. (You can call `./bin/civibuild create foo`... many CLI tools support this interaction.)
 * There will be some constituency which already runs their own mix of `drush` or `wp-cli` before installing buildkit. They'd hit subtle incompatibilities if (a) you force buildkit to use their version or (b) you force them to run buildkit's version. Diagnosing such subtle incompatibilities would be... quite unintuitive.

So I landed on this approach: the user owns their command-line (they can pick/choose/prioritize PATHs however they want), and `civibuild` owns its internal processing.

The user-story you describe is a bit different -- the user tries to influence `civibuild`'s internal processing. This should actually be possible without any patches, though. The file `app/civibuild.conf` allows an admin to control the environment for `civibuild`. Just edit `app/civibuild.conf` and append `export PATH=/path/to/my/tool:$PATH`. Maybe we should add some comments/suggestions for `PATH` in `app/civibuild.conf.tmpl`?

A few more thoughts:
 * For the custom `amp` scenario, I've used a different work-around during development (e.g. `cd ~/src/amp && php -dphar.readonly=0 `which box` build  &&  /bin/cp amp.phar ~/buildkit/bin/amp`), but I can see how your approach might be desirable for a long-running fork.
[...]

PR #371, opened 2017-10-26
add some comments/suggestions for PATH in app/civibuild.conf.tmpl

michaelmcandrew:
@totten, following on from #368 I added a comment suggestion to the civibuild.conf.tmpl.

In `bin/civibuild` the path is exported.

In my experiments I found that an export is not necessary in the config file: when I added `drush --version` to `civicrm-buildkit/app/config/drupal8-demo/download.sh`, and checked the output with the `PATH="$HOME/bin:$PATH"` commented/uncomments, the output reported the expected drush version.

I only bring that up here in case I am missing something and we should suggest that people export the path in the config.

### DIFF
--- a/app/civibuild.conf.tmpl
+++ b/app/civibuild.conf.tmpl
@@ -48,6 +48,7 @@
 ## Example: Update civibuild's $PATH
 
 # PATH="$HOME/src/amp/bin:$PATH"
+# export PATH
 
 ## The above example forces civibuild to use the amp executable found
 ## in $HOME/src/amp/bin rather than the default from buildkit's bin
--- a/app/civibuild.conf.tmpl
+++ b/app/civibuild.conf.tmpl
@@ -47,8 +47,8 @@
 #############################################################################
 ## Example: Update civibuild's $PATH
 
-# PATH="$HOME/bin:$PATH"
+# PATH="$HOME/src/amp/bin:$PATH"
 
-## This is useful if you want to point civibuild to specific versions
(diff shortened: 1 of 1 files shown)


totten:
Nice. :)

Regarding `export`... for anyone following along at home, `export` causes the variable to propagate to subcommands. Continuing the `amp` example, let's consider `civicrm/bin/setup.{sh,conf}`  -- which makes a call to `amp export` to figure the DB credentials.  

 * If `civibuild.conf` does `export PATH`, then the updated `PATH` would propagate down from `civibuild` to `setup.sh`. `setup.sh` would use the new/custom `amp`.
 * If `civibuild.conf` does *not* `export PATH`, then updates don't propagate. `setup.sh`would use the old `PATH` -- and the old/stock version of `amp`.

In this particular case, your build is probably OK because the *particular `amp` patches* deal with the vhost template -- and they're not relevant to `setup.sh` or `amp export`. So calling the stock or custom build is OK.

But... as a general rule-of-thumb for shell-scripting, I'd export `PATH` unless you have a reason not to. This is incongruous with the other variables, but unfortunately I think the incongruity is inevitable. (Some variables should really *not* be exported habitually, such as `ACTION` or `CACHE_TTL`, because they're at risk of unintentional collisions. Think of this has a programming language with multiple scopes -- and, for every variable, you target the appropriate scope.)

I'll tweak and merge the change.

michaelmcandrew:
Sorry :) Yes - that makes a lot of sense.

PR #373, opened 2017-11-06
add civicrm-admin-utilities to wp-demo

kcristiano:
Add Plugin for WP to demo-wp install type.   https://github.com/christianwach/civicrm-admin-utilities  WP repo at https://wordpress.org/plugins/civicrm-admin-utilities/   Add direct from wp repo via wp-cli to get the stable branch.

From the README:

- Modifies the styling of the CiviCRM menu to fix a number of issues
- Fixes the appearance of the WordPress Access Control form
- Adds a handy CiviCRM Shortcuts menu to the WordPress Admin Bar
- Allows you to choose which Post Types the CiviCRM shortcode button appears on
- In WordPress multisite, allows you to hide CiviCRM on sub-sites

This is a great addition to buildkit when using as both a demo site and for development.

### DIFF
--- a/app/config/wp-demo/install.sh
+++ b/app/config/wp-demo/install.sh
@@ -53,6 +53,7 @@ wp eval '$home = get_page_by_title("Welcome to CiviCRM with WordPress"); update_
 
 wp plugin activate civicrm
 wp plugin activate civicrm-demo-wp
+wp plugin install civicrm-admin-utilities && wp plugin activate civicrm-admin-utilities
 
 civicrm_apply_demo_defaults
 


kcristiano:
@totten this has evolved as an add-on @christianwach has maintained.   I do agree that a portion of this _could_ go in the main package, having a plugin allows iteration for WP at it's own pace.

Perhaps as we get dependent on this we'll add to core.  For now I'd like to get in the demo and then we can see what bits can be merged in.

christianwach:
@totten New code in WordPress is almost always developed via "feature plugins" before being merged into core. I'm sure I don't need to spell out why this is a *Good Thing™* :-) 

In the case of CAU, these fixes serve as examples of how to modify CiviCRM without touching core - hopefully a useful thing for devs who are fresh to the Civi-in-WP environment. Plus (if I'm honest) I'm too lazy to go through the *Jira > BuildKit > PR > QA > etc > etc* process when I can fix something for *all versions of CiviCRM* immediately.

Anyway, feel free to grab anything from CAU for CiviCRM core - I would just ask that you notify me so that I can amend CAU to take account of the changes.

kcristiano:
Thanks @christianwach I agree that iterating via plugin is the best way to go.   For now I want to add this to buildkit.   This will add the features to the default WP demo site as well.

totten:
> ...I'm too lazy to go through the Jira > BuildKit > PR > QA > etc > etc process...

Yeah, it feels like there are really common dev-stories where some of the steps are extraneous. We do need to improve the flow.

Definitely 👍 on "feature plugins" where substantive changes are made via add-ons. 

It might help to coin a phrase for contrast: a "bugfix plugin" would seem pretty weird to me:
 * (__Product perspective__) The main product should work. No one wants to ship things which are broken out-of-the-box, and no one wants to download things which are broken out-of-the-box.
 * (__Maintenance perspective__) *Fixes* tend to have *high coupling* to their context. (For example, CSS is implicitly coupled to a bunch of markup.) If the context changes (eg due to upstream changing; or due to multiple overlapping plugins), then the "bugfix plugin" is just as likely to cause problems, and the process of diagnosing and re-fixing is... eh... squiggly.
 * The act of *distributing* the bugfix plugin implies that it's an on-going concern.

But that's an *abstract* discussion.

In this particular case, `civicrm-admin-utilities` seems like it's *mostly* a "feature plugin". And it's great to have features in the demo sites. :)

christianwach:
>  a "bugfix plugin" would seem pretty weird to me

@totten Yes, I agree with you - except the fixes in CAU get applied to *all versions* of CiviCRM post-4.6, not just the latest release. This has been pretty handy IME for clients who cannot - for one reason or another - upgrade to the latest and greatest CiviCRM. Although zero-grading probably shouldn't be encouraged!

I have assigned CRM-20358 [WordPress access control table is basically unusable](https://issues.civicrm.org/jira/browse/CRM-20358) to myself since it's a genuine bug. IMO, the menu styling isn't really a "fix" as such - more of a design preference.

PR #374, opened 2017-11-09
Fix Yoti Version

mickadoo:
The default Yoti version (if unspecified) now relies on PHP 5.6 so we need to lock the version to 1.4, which supports PHP 5.5

This was already done in the hr16 build, but I forgot to add it to hr17, which this PR does.

### DIFF
--- a/app/config/hr17/drush.make.tmpl
+++ b/app/config/hr17/drush.make.tmpl
@@ -178,6 +178,9 @@ projects[masquerade][version] = "1.0-rc7"
 projects[logintoboggan][subdir] = civihr-contrib-required
 projects[logintoboggan][version] = "1.5"
 
+projects[yoti][subdir] = civihr-contrib-required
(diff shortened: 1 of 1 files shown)


JoeMurray:
@totten I'm a bit uncomfortable with our general tooling making reference to specific hardware that is not open source so far as I know. I'd even prefer if CiviHR did not end up with a dependence on use of specific hardware. Should this be refactored into optional extensions?

totten:
(1) I don't believe Yoti makes any hardware (aside from rebranded USB conference swag). Their main offering is an online service and mobile app -- which would put this in a similar category as PayPal or Authorize.net.

(2) `hr17` should be the demo configuration for CiviHR -- it doesn't impact dev/build processes for the main project.

JoeMurray:
Thanks, my mistake. I must have misunderstood something during Jamie's explanation of the service - I thought the swag was something special for holding private keys.

PR #376, opened 2017-11-21
app/config - Sync users during installation

totten:
Background
----------
When setting up the test bot for Mosaico's E2E tests, some tests failed
because they needed to lookup the admin user.

You can inspect the user/contact mappings by running a SQL query,
`select * from civicrm_uf_match;`

Before
------
In a new build, the admin and demo users exist. The demo user
has a corresponding `Contact` record, but the admin user doesn't.

After
-----
Both users have `Contact` records.

Comment
-------
So far, I've applied this on `test-ubu1204-1` and run the
Mosaico tests on top of the `drupal-clean` configuration. The
changes fixed that test.

However, I'm going to locally run several other build scripts to
ensure they don't regress.

### DIFF
--- a/app/config/drupal-case/install.sh
+++ b/app/config/drupal-case/install.sh
@@ -40,6 +40,7 @@ pushd "${WEB_ROOT}/sites/${DRUPAL_SITE_DIR}" >> /dev/null
     | drush cvapi setting.create --in=json
   ## Note: CiviGrant disabled by default. If you enable, update the permissions as well.
   civicrm_apply_demo_defaults
+  cv ev 'return CRM_Utils_System::synchronizeUsers();'
 
   ## Setup theme
   #above# drush -y en seven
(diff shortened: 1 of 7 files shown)


totten:
Tested now on `drupal-clean`, `drupal-demo`, `wp-demo`. Seems to work as expected.

I did notice a small change while testing -- before,  the user`demo` would be contact `202`; if you logged in as `admin`, then `admin` would be contact `203`. This PR syncs the users<->contacts very early, so the CID's are flipped.

This should be fine -- either mapping is legit. But sometimes tests might hard-code CIDs, so I did a grep on the codebase:

```
rgrep '202' | grep -v '202[0-9]' | grep php
```

Of those matches, only one actually seems to be a contact reference (`civicrm/tests/phpunit/CRM/Contribute/BAO/ContributionTest.php`). That test adds some sample-data for `contact_id` 202, but (in my local test run) it seems to work with either way. (To wit: the test doesn't really care who the user is; it just needs some `contact_id`.)

PR #385, opened 2017-12-16
Don't exclude DAO files from linting

colemanw:
Once https://github.com/civicrm/civicrm-core/pull/11418 is merged we should merge this as well to lock in the testing of these files.

### DIFF
--- a/bin/civilint
+++ b/bin/civilint
@@ -51,7 +51,7 @@ function split_files() {
     if [ ! -e "$file" ]; then
       echo "$file" >> "$skipfiles"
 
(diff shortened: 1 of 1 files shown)


totten:
Perhaps we should merge this about a week *after* the DAO files in `civicrm-core#master` become compliant?

colemanw:
Maybe a whole release cycle afterwards. But yea.

MikeyMJCO:
So this was merged for core in 4.7.31 - is this PR ready to go into buildkit?

PR #386, opened 2018-02-05
upgrade-test - Make version checks more adaptable

totten:
https://lab.civicrm.org/development-team/Release-Management/issues/1 will realign the version numbering from the third digit (`4.7.{$z}`) to the second digit (`5.{$y}.0`).

However, some parts of the upgrade-testing assumed that the second digit doesn't change very often -- they are hardcoded to the given value of the second digit.

With this revision, the upgrade-testing should provide similar behavior without requiring the hard-coded values.

### DIFF
--- a/bin/civi-test-run
+++ b/bin/civi-test-run
@@ -81,16 +81,7 @@ function task_karma() {
 function task_upgrade() {
   ## Run the tests -- DB upgrade tests
   CIVIVER=$(getCiviVer)
-  case "$CIVIVER" in
-    4.3*)    UTVERS="4.2.9-multilingual_af_bg_en.mysql.bz2 4.3.0*" ;;
-    4.4*)    UTVERS="4.2.9-multilingual_af_bg_en.mysql.bz2 4.3.0* 4.4.0*" ;;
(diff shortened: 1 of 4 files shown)


totten:
Hey, @seamuslee001 -- you might be interested in the updates to `civi-test-run`.

seamuslee001:
@totten looks good, we may want to update the dev docs in correspondence. Code changes look sane here.

totten:
OK, thinking a bit more about how/why one would describe this update at that level...

The real issue is that running all the upgrade tests on PRs might take too long, and you still get a lot of utility from a smoke-test with a couple diverse data-sets. That phenomenon -- wanting to run a quick smoke-test instead of a full test -- can be applied to most suites.

So thinking outloud -- is it better to:
 * Introduce a general notation for `civi-test-run` to say "give me the quick smoketest for suite X". (Pro: Easier to learn `civi-test-run` options. For PR testing on `civicrm-{core,drupal,wordpress,packages}`, it's a pretty suitable option. Con: "smoketest" is a bit abstract and will encompass policy opinion, and that will inevitably cause friction for someone.)
 * Proceed as written -- `civi-test-run upgrade` supports a special notation for specifying a smoketest.

totten:
Updated `civicrm-upgrade-test`. Rebased.

It would be nice to have more support for smoke-testing in `civi-test-run`, but this is fine for now.

PR #390, opened 2018-02-12
Use more recent phpunit with support for 5.3, 5.4 php gone

eileenmcnaughton:


seamuslee001:
I don't believe this can happen until 4.6 goes EOL because 4.6 requires phpunit3, I believe tim did much work as part of testapaloza that meant that 4.7 and up could run on phpunit4

eileenmcnaughton:
@seamuslee001 does t require phpunit 3 because of the php versions or other - is there some agreement from @jackrabbithanna that could allow this to be merged - there is an ongoing time cost in using out of date phpunit versions - even just in the extra support that has to be given to other people

eileenmcnaughton:
Alternatively could we fork buildkit with a 4.6 branch that is never touched again & progress the main branch ?

eileenmcnaughton:
@seamuslee001 so to be really clear what is the impact if we merge this? What will break ?

Buildkit affects (AFAIK)
- local developer experience (I think as long as there is a version that works on 4.6 we can tag a release & then only dev for 4.7)
- how the CI boxes operate - this seems a bigger deal - ie. the LTS team has commited the Core Team to supporting 4.6 until the end of this year. However, LTS team has agreed we no longer need to test on php 5.4

seamuslee001:
At present 4.6 uses an embedded phpunit version 3 for its testing. Which is contained in packages folder. This means we need phpunit3 phar to run it

in this PR we are telling composer to download phpunit4 rather than phpunit3. for 4.6 to work off phpunit4 we would need some changes like https://github.com/civicrm/civicrm-core/pull/12079 to make it work i think. 

so it would break by likely killing 4.6 unit tests at present. 

To run the unit tests on php7.2 phpunit4 has incompatabilities that mean virtually impossible but phpunit5 in my local testing is showing it works my alternate pr here #405 means that we would download phpunit5 (need a core commit on 5.x to tell it to use it) but that is all that would be required for master to then run on php7.2 in a CI context.

seamuslee001:
Note that the fact we can't run phpunit4 on php7.2 is immaterial to this issue, the issue here is that we would likely break 4.6 testing because phpunit3 would disappear.

eileenmcnaughton:
@seamuslee001 to clarify - this means - if we took 4.6 out of the equation then it would be fairly trivial to do CI against php 7.2 on master (by upgrading phpunit).

But if we are supporting php 4.6 then either
a) significant work would go into updating php 4.6 to run on a later version - it looks like you have also done a chunk of work towards that but my guess is that it would still take a significant chunk more (by 'someone' or 'someones')
b) we can't support php 7.2 until 4.6 support ends (Dec 2018)
c) we should get the funds 'from somewhere' for a second test box for 4.6 so it can run on the 'old' version of buildkit & we can upgrade our main box to run a version of phpunit that supports 7.2

eileenmcnaughton:
(ie there are 2 key reasons to update phpunit 
1) to support php 7.2
2) to reduce the ongoing time costs (to many people) of working with an older phpunit version that is likely different to the one they use for other unit tests on their system )

seamuslee001:
@eileenmcnaughton yeh tho it might be interesting to do a local test and see if 4.6 does actually break after running composer install here. (btw looking at the test run it may not actually require much more to get 4.6 on phpunit4)

eileenmcnaughton:
@seamuslee001 OK sure - I guess we wouldn't want to sink more than a couple of hours of 'discovery' on this though given that even if you get it working there is a review task of at least equal time that no-one has committed to doing yet

seamuslee001:
@eileenmcnaughton @totten I take back my comments re 4.6 i just ran a test of upgrading phpunit to v5 on a php5.6 box i have and then trying to run unit tests for 4.6 and they worked.

I would suggest that we should be upgrading phpunit in composer to v5 but leaving the phpunit4 download in just as a backup

eileenmcnaughton:
Thanks for putting in the time to test what is involved in continuing to support 4.6 @seamuslee001  - @totten I vote to update to phpunit v5 what do you think?

composer.json
```diff
@@ -34,6 +34,10 @@
     {
       "type": "git",
       "url": "https://github.com/totten/paratest.git"
+    },
+    {
+      "type": "git",
+      "url": "https://github.com/eileenmcnaughton/php-symbol-diff.git"
```
totten: Thanks for tracing this. I think it's resolved after committing https://github.com/totten/php-symbol-diff/commit/8d1a1e31867275583c75c15896b6a8cf0b540439

composer.json
```diff
@@ -15,13 +15,13 @@
...
-    "php": ">=5.3.3",
+    "php": ">=5.5",
     "drush/drush": "dev-master#d1d13676f5beacaa9f8619088fe3ae45ea28a6cf",
-    "totten/php-symbol-diff": "dev-master#54f869ca68a3cd75f3386f8490870369733d2c23",
+    "totten/php-symbol-diff": "dev-master",
     "civicrm/upgrade-test": "dev-master#b93207fd28ac90426e9c95d572068f2e907b08e2",
     "drupal/coder": "dev-8.x-2.x-civi#a801890a82d52e23a83c5d820270b6e228843d79",
     "brianium/paratest": "dev-batching as 0.7",
-    "phpunit/phpunit": "3.7.27",
+    "phpunit/phpunit": ">=4",
```
totten: I want to clarify a little bit about what this means, because it's probably confusing.

Buildkit includes two copies of `phpunit` -- one via `composer.json` (v3) and one via PHAR (v4).  This patch changes the `composer.json` copy, which isn't actually used by any of our CI today. It may be used locally by some developers (because the name is more familiar), but I have no way to gauge this.

My gut reaction is that we should (1) remove phpunit from buildkit's `composer.json` and (2) optionally add a stub script `bin/phpunit` which calls `bin/phpunit4` or `bin/phpunit5` .

There is a longer discussion to be had about the general framing of per-project tooling vs global tooling (and how neither is really prefect for our ecosystem). But in terms of keeping changes small+manageable, I'd vote to remove `phpunit` from this `composer.json`.

seamuslee001: I would be ok with that as well

eileenmcnaughton: I'm happy with that - anything that gets rid of php 3.7!

I still don't understand why we have this issue per se - is it just a vestige of prior support for php 5.4 (as I originally assumed) or is there some remaining complexity in our CI env?

totten: OK, closing this. Replaced by #408.

PR #391, opened 2018-02-12
External build types

michaelmcandrew:
Adding an option --type-dir that allows you to specify the directory that you want to look in for build files


totten:
@michaelmcandrew I think this is generally well-formed, and it manages to be consistent with the bash-coding style -- which is quite a feat!

Before merging, I want to check with you about a design question -- how would you compare the user-experience of these two approaches:

* Add a CLI option  to *specify* the type-dir. If you want register a new type outside the `civicrm-buildkit.git`, then pass `--type-dir /path/to/my/location` whenever you call `civibuild`.
* Run a *search* to locate the type-dir (e.g. check `/etc/civibuild/config/$TYPE`; then check `$PRJDIR/app/config/$TYPE`). Use the first match. (Optionally, control this with an env-var like `export CIVIBUILD_CONFIGS=/srv/myconfigs:/etc/civibuild/config:/srv/buildkit/app/config`.)

It seems like the CLI option is easier to discover (via `civibuild create -h`), but the search might be more usable on a day-to-day basis (because you don't have to remember/type the extra args)?

(*Aside: In both cases, the implementation complexity should be similar -- mostly just updating `civibuild.compute-defaults.sh` to compute `SITE_CONFIG_DIR`.*)

michaelmcandrew:
Hey there, I'm going to try and understand and rephrase your questions...

I think there are two (independent) parts to your question.

1. Choose between
a) allowing people to override the single directory that is searched for types
b) allowing people to specify a list of directories that can be searched in order

2. Allow people to set the above via an environment variable as well as via a command line option

I would be up for 1a or 1b and would also be up for 2.

1b and 2 could be implemented after 1a without breaking backward compatibility.

So, if I undestand you correctly, yes to everything :)

totten:
Valid point -- the env var and CLI option are not mutually exclusive. 

I think 1b+2 would be the ideal.

Just for kicks, here's a bit of bash code (inspired by some googling) that may help with searching a colon-delimited list:

```bash
## Search a colon-delimited list of folders for a file.
## Ex: findItem whoami "/usr/local/bin:/usr/bin:/bin"
function findItem() {
  local SEARCH_TARGET=$1
  local SEARCH_PATHS=$2
  printf %s "$SEARCH_PATHS" | awk 'BEGIN {RS=":"}; 1' | while read DIR ; do
    if [ -e "$DIR/$SEARCH_TARGET" ]; then
      echo "I found '$SEARCH_TARGET'! It's really "$DIR/$SEARCH_TARGET"!"
      break
    fi
  done
}

findItem whoami "/usr/local/bin:/usr/bin:/bin"
```

michaelmcandrew:
Looks nice.

We could always merge this now and add a seperate issue for the multiple search paths at a later date, since they are not mutually exclusive approaches (guidance in a future version would be along the lines of

> --type-dir - Can be a single directory or colon seperated list of directories to search in order.

totten:
Replaced with #471, which should allow the same use-case:

```
## With 391
civibuild create foo --type-dir /home/foo/myconfigs

## With 471
CIVIBUILD_PATH=/home/foo/myconfigs civibuild create foo
```

PR #392, opened 2018-02-14
Skip CodeGen generated files

monishdeb:
Skip install/langs.php and CRM/Core/I18n/SchemaStructure.php files from civilint check

### DIFF
--- a/bin/civilint
+++ b/bin/civilint
@@ -62,6 +62,9 @@ function split_files() {
         echo "$file" >> "$skipfiles"
       fi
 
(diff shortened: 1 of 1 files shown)


monishdeb:
Result : 
```
civilint install/langs.php CRM/Core/I18n/SchemaStructure.php 
===========================[ Identify Files ]===========================
Skip Files:
 * install/langs.php
 * CRM/Core/I18n/SchemaStructure.php
```

totten:
Tested locally. These two files are no skipped (as expected), but other files generally still work (as expected). Thanks. 👍

PR #404, opened 2018-05-02
Add in support for ubuntu bionic

seamuslee001:


totten:
Good idea. Seamus did some testing on this for bionic. Also, it doesn't pose any real risk other/existing distros.

PR #408, opened 2018-05-22
composer.json - Remove extra copy of phpunit (et al)

totten:
CiviCRM's usage of PHPUnit evolved along this track:

1. In `civicrm-core` and `civicrm-packages`, there was a forked copy of PHPUnit 3.x. It had various patches applied,
   and the test suite didn't run with vanilla PHPUnit. (Things ran this way for several years.)
2. Added vanilla phpunit 3.x to `civicrm-buildkit` (via `composer`). Patched `civicrm-core` until it's compatible.
   (Things didn't run this way as long -- maybe a few months?)
3. Added vanilla phpunit 4.x to `civicrm-buildkit` (via `phar`). Patched `civicrm-core` until it's compatible.
   (Things have now run this way a couple years.)

The take-away is that the composer-based copy of phpunit 3.x hasn't been used
for a long time.  (Not by our CI, anyway.) There's been some confusion about
this (e.g.  because the old version has the default command name `phpunit`).

Following the [discussions on #390 and #345](https://github.com/civicrm/civicrm-buildkit/pull/390#discussion_r186237206), I'm submitting a replacement PR which removes the unused PHPUnit 3.x. It also removes a few related packages which are also unused in `civicrm.org` CI.

Note: After removing this copy of phpunit on my local, I successfully ran
the test suites for 4.6 and 5.x (to confirm the expectation that it was
unused).


eileenmcnaughton:
I've just upgraded to this & I 'feel' like I'm doing as you describe - but I'm missing PHPUnit_Extensions_Database_TestCase - per this https://stackoverflow.com/questions/23986137/fatal-error-class-phpunit-extensions-database-testcase-not-found
when using the phpunit5.phar file (am now in the process of switching to not using that & using my own composer version)

composer require --dev phpunit/dbunit

totten:
> I note your example does an include on the vendor dir but the dbunit & selenium have come out of there

The `vendor` reference there is unnecessary. `phpunit4.phar` and `phpunit5.phar` both include copies of `PHPUnit_Extensions_Database_TestCase`. I removed `vendor` from my list of includes and restarted PHPStorm -- and things still worked fine:

* A couple arbitrarily chosen tests (`api_v3_ContactTypeTest` and ` CRM_Utils_Cache_SqlGroupTest`) are working.
* If I open `CiviUnitTestCase` and `Cmd-Click` on `PHPUnit_Extensions_Database_TestCase`, then it pulls up the code. The breadcrumb bar shows the path as `phpunit4.phar > dbunit > Extensions > Database > TestCase.php`. (And I get similar results if the IDE is configured for `buildkit/extern/phpunit5` instead of `buildkit/extern/phpunit4`.)
* If I inspect `phpunit4.phar` or `phpunit5.phar` via CLI, I can find the file:

```
$ phar list -f ~/bknix/civicrm-buildkit/extern/phpunit4/phpunit4.phar  | grep Database/TestCase
|     |-phar:///Users/me/bknix/civicrm-buildkit/extern/phpunit4/phpunit4.phar/dbunit/Extensions/Database/TestCase.php

$ phar list -f ~/bknix/civicrm-buildkit/extern/phpunit5/phpunit5.phar  | grep Database/TestCase
|     |-phar:///Users/me/bknix/civicrm-buildkit/extern/phpunit5/phpunit5.phar/dbunit/Extensions/Database/TestCase.php
|     |-phar:///Users/me/bknix/civicrm-buildkit/extern/phpunit5/phpunit5.phar/dbunit/Extensions/Database/TestCaseTrait.php

$ md5sum ~/bknix/civicrm-buildkit/extern/phpunit*/phpunit*.phar
f8711c03e09d6ccaa51586b6d5eabc7d  /Users/me/bknix/civicrm-buildkit/extern/phpunit4/phpunit4.phar
a2a8f69a743bd5650a35b96fc8b85328  /Users/me/bknix/civicrm-buildkit/extern/phpunit5/phpunit5.phar
```

Here's the current set of config screens on my workstation:

screen shot 2018-06-30 at 1 17 33 pm
screen shot 2018-06-30 at 1 17 51 pm
screen shot 2018-06-30 at 1 18 24 pm
screen shot 2018-06-30 at 1 20 46 pm

(Note: The `PATH` shouldn't be necessary for most folks, but I have to use it because some of my binaries in weird places.)

PR #412, opened 2018-06-13
Load civirules from lab.civicrm.org

jaapjansma:
We have moved CiviRules from github to lab.civicrm.org
Buildkit should fetch civirules from this new location.


jaapjansma:
Thaks.

PR #414, opened 2018-06-20
Create 400-error-handling.php

eileenmcnaughton:
Turn on all possible noise

### DIFF
--- a/app/civicrm.settings.d/400-error-handling.php
+++ b/app/civicrm.settings.d/400-error-handling.php
@@ -1,0 +1,1 @@
+ini_set('error_reporting', E_ALL);


totten:
@eileen @seamuslee I generally like the policy of using `E_ALL` in development. Let me verbalize some thoughts that came up with this. (They're not all in agreement -- more like a brain-dump.)

* The `www-demo` host probably shouldn't use `E_ALL`. We've already gotten comments about it showing too much debug info. We can do a local override, so that's OK.
* The file in this PR isn't syntactically right -- it needs to start with `<?php`. But if that's the only thing, I can clean it up while merging.
* I like this for how it would affect the default local developer environment. However, I'm nervous about putting it into CI -- would rather see the test-suites passing under a more rigorous policy before it goes live there. Some data points on CI:
    * `test-ubu1204-{1,5}` currently default to `E_ALL & ~E_DEPRECATED`.
    * `test-ubu1604-1` currently defaults to `E_ALL & ~E_DEPRECATED & ~E_STRICT`.
    * `CiviUnitTestCase::__construct()` sets the reporting level to `E_ALL & ~E_NOTICE`.
* I'm not sure `CiviUnitTestCase.php` is a great place to put the call to `error_reporting(...)`. The nice thing about putting the policy in a place like `400-error-handling.php` is that it can be overridden. Allowing overrides means that you can, e.g., have two buildbots giving feedback at different levels of strictness.

A few ideas...

* We could just deploy this locally on `test-ubu1204-1` and `test-ubu1604-1` and see if any problems show up in the matrix. (That way problems don't hurt PR process.) And if that works, then merge this in as a general policy.
* We could open a PR which patches `CiviUnitTestCase` to be more aggressive about `error_reporting(E_ALL)`. I'm not suggesting we merge it now -- but it would be interesting to have a ballpark for close or far we are from passing at a stricter level.

eileenmcnaughton:
@totten I agree we should see what this does in CI!

eileenmcnaughton:
Let's kill based on stale tag

PR #415, opened 2018-06-29
PCHR-3849: Google Analytics Implementation - Milestone 1

AkA84:
Contains the following PRs:
* #3

### DIFF
--- a/app/config/hr17/drush.make.tmpl
+++ b/app/config/hr17/drush.make.tmpl
@@ -222,6 +222,12 @@ projects[menu_attributes][version] = 1.0
 projects[roles_for_menu][subdir] = civihr-contrib-required
 projects[roles_for_menu][version] = 1.1
 
+projects[google_tag][subdir] = civihr-contrib-required
+projects[google_tag][version] = 1.3
+
+projects[datalayer][subdir] = civihr-contrib-required
+projects[datalayer][version] = 1.1
+
 ; ****************************************
 ; Compucorp custom drupal modules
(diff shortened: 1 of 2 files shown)


JoeMurray:
Some of us are concerned about privacy. I just want to confirm that Google Analytics can be turned off.

ginkgomzd:
I would urge transparency on this issue. "Google Analytics" could be perceived as disingenuous, when in fact what is included is Google Tags. 
https://www.drupal.org/project/google_tag
These do have very significant privacy concerns, as any third-party code could be injected into the site, not just code from google. (I'm not saying anyone could inject it, just that the account admin could inject any kind of code.)

I have not seen anything concerning, but again, I urge transparency, and heading off any scares with a blog post sounds like a good idea.

totten:
Since I haven't worked with those modules, I haven't formed an opinion about them and appreciate some of the insight. 

It may be worth pointing out that this patch applies to the `hr17` build, which is essentially a reference/demo build for CiviHR. So it has no implications for the dependencies of a typical Civi installation or for `civicrm.org`'s normal test/CI builds (`drupal-demo`, `drupal-clean`, `wp-demo`, etc).

I view `hr17` as more akin to the various [demo sites](http://civicrm.org/demo) where partners/collaborators exercise their own judgment about what "mix" to put together. Actually, come to think of it, I don't know if those demo configurations are published --  `hr17` feels pretty transparent because you can see the build-scripts here. So kudos to HR folks on being ahead of the curve.

PR #416, opened 2018-06-29
For build types which install CiviVolunteer…

GinkgoFJG:
…set the default branch for the org.civicrm.volunteer and org.civicrm.angularprofiles repos to master.

Removed reference to CiviVolunteer from wp-empty, which doesn't install CiviCRM at all.


totten:
Thanks, @GinkgoFJG. I'm inclined to take your word on this one. :)

PR #417, opened 2018-07-03
infrastructure/ops#580: Avoid error popups on demo site login.

mlutfy:


seamuslee001:
change looks good to me @mlutfy @totten

MegaphoneJon:
Same here @mlutfy!

MegaphoneJon:
Can we implement the same change on `wp-demo`?

mlutfy:
Good idea. Done.

mlutfy:
I'm taking the liberty to merge this, based on

- feedback received
- it's fairly trivial
- it mostly affects infra and I have to deal with it.

mlutfy:
I rebuilt dmaster via Jenkins on www-demo, and seems OK.

totten:
I was concerned about shushing warnings on dev sites. This patch is reasonable given the split between `drupal-demo` and `drupal-clean`, ie

* Demo sites use `drupal-demo`, which shushes some warnings.
* Dev sites use `drupal-clean`, which doesn't shush those warnings.

That's not a perfect rationalization... e.g. local dev builds often use `drupal-demo` or `wp-demo`. In the long run, I'd like to have a way to make a per-server declarative policy -- e.g. it'd be great if one we could create a file on `www-demo` in `/etc/civicrm.settings.d/300-status-overrides.php` with something like

```php
global $civicrm_status;
$civicrm_status['checkOutboundMail']['ignore_severity'] = 'critical';
```

The current patch is fine interim measure.

PR #421, opened 2018-08-07
Fix error install CiviCRM demo instances with CiviCRM 5.x

mattwire:
The version check for 4.7.0 gives a false positive for 5.x and civicrm_apply_demo_defaults fails.  This fixes the issue.

### DIFF
--- a/src/civibuild.lib.sh
+++ b/src/civibuild.lib.sh
@@ -523,7 +523,7 @@ EOSQL
 ###############################################################################
 ## Appy more default values
 function civicrm_apply_demo_defaults() {
-  if cv ev 'exit(version_compare(CRM_Utils_System::version(), "4.7.0", "<") ?0:1);' ; then
+  if cv ev 'exit(version_compare(CRM_Utils_System::version(), "4.7", "<") ?0:1);' ; then
     cv api setting.create versionCheck=0 debug=1
   fi
   cv api MailSettings.create id=1 is_default=1 domain=example.org debug=1


src/civibuild.lib.sh
```diff
@@ -523,7 +523,7 @@ EOSQL
 ###############################################################################
 ## Appy more default values
 function civicrm_apply_demo_defaults() {
-  if cv ev 'exit(version_compare(CRM_Utils_System::version(), "4.7.0", "<") ?0:1);' ; then
+  if cv ev 'exit(version_compare(CRM_Utils_System::version(), "4.7", "<") ?0:1);' ; then
```
totten: @mattwire OK, I have to admit to two points of confusion -- not sure what the original problem is, and not sure how this patch changes the behavior.

My understanding is that the `versionCheck` setting exists in 4.6 but not in 4.7/5.x, so the code tries to omit `versionCheck` on newer builds. The description seems to suggest that the `version_compare()` would misbehave in 5.x, so I tried using that function with a range of inputs:

```php
$examples = [];
foreach (['4.6.20', '4.7.0', '4.7.1', '4.7.20', '5.0.alpha1', '5.0', '5.0.0', '5.1.beta1', '5.2.2'] as $v) {
  $examples[$v] = version_compare($v, "4.7.0", "<");
}
var_export($examples);
```

which yields

```php

array (
  '4.6.20' => true,
  '4.7.0' => false,
  '4.7.1' => false,
  '4.7.20' => false,
  '5.0.alpha1' => false,
  '5.0' => false,
  '5.0.0' => false,
  '5.1.beta1' => false,
  '5.2.2' => false,
)
```

and that looks right to me.

mattwire:
Right, well if we can't reproduce it and I can't remember we'll close it

PR #423, opened 2018-08-27
Issue 422-Prepare Buildkit for testing Gutenberg on WordPress

kcristiano:
WordPress 5.0 will be released this year.  We can guess when that will be but it should be between Late September and December 1st based on information I have.

We will need to have Buildkit prepped for testing.

- Update WP to latest stable version
- Update wp-cli to latest stable version
- Install and activate the plugin Gutenberg Ramp to allow demo sites to be built _without_ Gutenberg
- Add the Gutenberg Plugin but do not activate

related https://lab.civicrm.org/dev/wordpress/issues/3


app/config/wp-demo/install.sh
```diff
@@ -53,7 +53,9 @@ wp eval '$home = get_page_by_title("Welcome to CiviCRM with WordPress"); update_
 
 wp plugin activate civicrm
 wp plugin activate civicrm-demo-wp
-wp plugin install civicrm-admin-utilities && wp plugin activate civicrm-admin-utilities
+wp plugin install civicrm-admin-utilities --activate
+wp plugin install gutenberg
```
colemanw: Will this line break anything when 5.0 is released and Gutenberg isn't an extension anymore?

kcristiano: They say line 57 should not break once it's no longer a plugin.  We at most should get a 'plugin not found' message.  I added simply for ease of use.  I imagine that we may have to remove when 5.0 is released.

We could remove now and have people manually install.   I am flexible.  I did think having it installed but not active would be better and encourage testing with it.

PR #425, opened 2018-09-22
Issue #394: Use .test, not .dev

### LINKED ISSUE #394: .dev sites on google chrome, etc.
Originally from https://github.com/civicrm/civicrm-buildkit-vagrant/issues/32

> 
> This is a minor quality of life issue.
> 
> Since version 66 (Dec 2017) Chrome now force redirects all sites ending in the .dev domain to https even on the localhost. see this write up on the issue.
> 
> This causes a server response error when trying to view the pre-generated sites on chrome (Other browsers work fine).
> 
> We can manually add/modify existing host setting but it if it is not too large a job it may be worth modifying the .dev to something else like .devel or .localhost in the default settings.
> 
> Again thank you so much for all the work on this project!
> 
>

xurizaemon:
Fixes #394 

See also civicrm/civicrm-buildkit-vagrant#40

### DIFF
--- a/src/civibuild.aliases.sh
+++ b/src/civibuild.aliases.sh
@@ -96,6 +96,6 @@ function civibuild_alias_resolve() {
   esac
 
 #  if [ -n "$known" -a "%AUTO%" = "$URL_TEMPLATE" ]; then
-#    URL_TEMPLATE='http://%SITE_NAME%.dev'
+#    URL_TEMPLATE='http://%SITE_NAME%.test'
 #  fi
 }
--- a/src/civibuild.compute-defaults.sh
+++ b/src/civibuild.compute-defaults.sh
@@ -16,7 +16,7 @@
 if [ -z "$CMS_URL" ]; then
   if [ "%AUTO%" == "$URL_TEMPLATE" ]; then
     if [ -n "$IS_ALIAS" ]; then
-      CMS_URL=$( echo "http://%SITE_NAME%.dev" | sed "s;%SITE_NAME%;$SITE_NAME;g" )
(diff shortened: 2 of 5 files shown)


xurizaemon:
Tested civicrm/civicrm-buildkit#425 inside civicrm/civicrm-buildkit-vagrant#41. 

`civibuild create d7-master` creates http://d7-master.test and behaves correctly.

MikeyMJCO:
Looks good to me - tested locally and it works fine. Avoids the horrible HSTS issue xD

totten:
Alright, the patch looks generally as-advertised, and I'm inclined to trust @MikeyMJCO's testing. 

This leaves us with the communication/migration issue. Personally, I use both`dnsmasq` and `mod_vhost_alias` in dev systems, and I've known others to do the same. It seems pretty plausible to me that someone (who relied on the default `*.dev` URL template) would find that `civibuild` no longer creates working builds. So...

I'm thinking to send a PSA to `civicrm-dev@` with this link: https://gist.github.com/totten/cda9a6ec1f03020ebc2fd3ca4434152b

@xurizaemon @MikeyMJCO @michaelmcandrew, if the notes/transition advice look OK to you, then I'll merge and send the notification. (Note: We could do it in the other order if we want to be really polite and don't mind waiting a little longer.)

xurizaemon:
Advice looks good.

, I can't think of how anyone could be using .dev today, unless they use ancient / freaky browsers or have gone to great effort to bypass the HSTS.

But, those people might exist, and the advice might help them, so :thumbsup:

MikeyMJCO:
Advice looks good. 

No further ".dev" mentions in code. No docs to update (.dev isn't mentioned in the developer docs).

PR #429, opened 2018-09-29
Bump Drush8 to 8.1.17

MikeyMJCO:
Simple bump of the Drush version for better PHP7.2 support.

[Drush 8.1.17 release notes](https://github.com/drush-ops/drush/releases/tag/8.1.17)


totten:
Good idea 👍

MikeyMJCO:
@totten Looks like commit https://github.com/civicrm/civicrm-buildkit/pull/429/commits/76ebb2f5c8ea21439b8048a729cf7413728b3688 snuck in from a dev branch. Probably best to revert that one for now - it's a half-finished commit from some debugging. Sorry!

totten:
Yeah, 76ebb2f had seemed a little out-of-place, but it looked safe enough, and it still works. If your debugging leads to something else, then cool. But it's not hurting anything.

PR #430, opened 2018-09-29
{drupal,wp,backdrop}-demo - Download and enable api4 on demo builds

totten:


totten:
Ping @colemanw

totten:
In retrospect, I probably shouldn't add the `cv en api4` to each `install.sh` -- it'd be cleaner to put it under `civicrm_apply_demo_defaults()`.

totten:
Per above, updated `civicrm_apply_demo_defaults()` to call `cv en api4`. This includes the option `--ignore-missing` so that it continues without error if any custom/downstream builds use this function without including api4.

PR #431, opened 2018-09-30
Ubuntu 18.04.1 Bionic Beaver - Final Fixes

MikeyMJCO:
Remove duplicate apt update. Wrap mcrypt installation in conditional 
logic to prevent an error if installed. Fix escaping of node warnings.
Fix system detection message on Debian Stretch.

### DIFF
--- a/bin/civi-download-tools
+++ b/bin/civi-download-tools
@@ -458,16 +458,12 @@ function do_system_bionic() {
     echo "Warning: This system will not support php-mysql extension"
     echo ""
     if cvutil_confirm "Run automated installation? [Y/n] " y y; then
-      sudo curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
-      sudo apt-get update
+      sudo curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash - # This runs sudo apt update so there's no point running it twice.
       sudo apt-get -y install $PACKAGES
       sudo phpenmod bcmath
       sudo phpenmod curl
       sudo phpenmod imap
       sudo phpenmod xml
-      sudo pecl install mcrypt-1.0.1
-      sudo bash -c "echo extension=/usr/lib/php/20170718/mcrypt.so > /etc/php/7.2/cli/conf.d/mcrypt.ini"
-      sudo bash -c "echo extension=/usr/lib/php/20170718/mcrypt.so > /etc/php/7.2/apache2/conf.d/mcrypt.ini"
       sudo a2enmod rewrite
       sudo apache2ctl restart
     else
@@ -737,7 +733,7 @@ check_php_ext hash recommended
 check_php_ext imap recommended
 check_php_ext intl recommended
 check_php_ext mbstring recommended
-check_php_ext mcrypt recommended
+check_php_ext mcrypt recommended "WARNING: Failed to locate the PHP \"mcrypt\" extension. Please install this manually to avoid issues with SMTP passwords or older versions of CiviCRM."
 if php -r 'exit(version_compare(PHP_VERSION, 7.0, "<") ? 0 : 1);' ; then
   check_php_ext mysql recommended  ## In case you try to install older Civi versions (<= 4.7.11)
 fi
@@ -747,8 +743,8 @@ check_php_ext session recommended
 check_php_ext soap recommended
 check_php_ext zip recommended
 
-check_command node recommended "WARNING: Failed to locate command 'node'. NodeJS (http://nodejs.org/) is required for development of CiviCRM v4.6+."
-check_command npm recommended "WARNING: Failed to locate command 'npm'. NodeJS (http://nodejs.org/) is required for development of CiviCRM v4.6+."
+check_command node recommended "WARNING: Failed to locate command \"node\". NodeJS (http://nodejs.org/) is required for development of CiviCRM v4.6+."
+check_command npm recommended "WARNING: Failed to locate command \"npm\". NodeJS (http://nodejs.org/) is required for development of CiviCRM v4.6+."
 
 ##################################################
 ## Cleanup


MikeyMJCO:
The NodeJS warnings at the bottom may be over-escaped but I used bash's own functions to get them escaped to a point which worked with `printf "%q" "STRING"`. 

All tested and working - including the mcrypt logic - no longer errors out if mcrypt is installed.

Can be tested using:
`curl -Ls https://github.com/MJCO/civicrm-buildkit/raw/ubuntu_bionic/bin/civi-download-tools | bash -s -- --full --dir ~/buildkit`

bin/civi-download-tools
```diff
@@ -458,14 +458,22 @@ function do_system_bionic() {
...
-      sudo curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
-      sudo apt-get update
+      sudo curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash - # This runs sudo apt update so there's no point running it twice.
       sudo apt-get -y install $PACKAGES
       sudo phpenmod bcmath
       sudo phpenmod curl
       sudo phpenmod imap
       sudo phpenmod xml
-      sudo pecl install mcrypt-1.0.1
+      if pecl list | grep mcrypt >/dev/null 2>&1 ; then
```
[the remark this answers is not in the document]
totten: This check seems like a good idea. :+1:

bin/civi-download-tools
```diff
@@ -502,7 +510,7 @@ function do_system_jessie() {
 function do_system_stretch() {
   set -e
     PACKAGES="acl git wget unzip zip mysql-server mysql-client php-cli php-imap php-ldap php-curl php-mysql php-intl php-gd php-mcrypt php-apcu apache2 libapache2-mod-php php-bcmath php-mbstring php-soap php-zip"
-    echo "Detected \"Debian Jessie\"."
+    echo "Detected \"Debian Stretch\"."
     echo ""
     echo "Installing nodejs Debian repository."
     wget -O- https://deb.nodesource.com/setup_8.x | sudo -E bash -
```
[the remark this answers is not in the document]
MikeyMJCO: The duplicate apt-get update is removed in PR #433

bin/civi-download-tools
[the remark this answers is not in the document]
MikeyMJCO: I haven't come across anything to indicate it'll be needed for any other versions of Ubuntu (which are using apt-get to manage mcrypt) or Debian (also using apt-get)

bin/civi-download-tools
```diff
@@ -747,8 +755,8 @@ check_php_ext session recommended
 check_php_ext soap recommended
 check_php_ext zip recommended
 
-check_command node recommended "WARNING: Failed to locate command 'node'. NodeJS (http://nodejs.org/) is required for development of CiviCRM v4.6+."
-check_command npm recommended "WARNING: Failed to locate command 'npm'. NodeJS (http://nodejs.org/) is required for development of CiviCRM v4.6+."
+check_command node recommended "WARNING:\ Failed\ to\ locate\ command\ \"node\".\ NodeJS\ \(http://nodejs.org/\)\ is\ required\ for\ development\ of\ CiviCRM\ v4.6+."
+check_command npm recommended "WARNING:\ Failed\ to\ locate\ command\ \"npm\".\ NodeJS\ \(http://nodejs.org/\)\ is\ required\ for\ development\ of\ CiviCRM\ v4.6+."
```
totten: Do you have any sense for why this encoding changed? e.g. Bionic has a newer version of `bash` which makes a subtle change in syntax?

The question is partially idle/self-education, but it's also trying to double-check that we're not getting in boxed-in -- e.g. if there's a compatibility-break in the `bash` syntax, then fixing for a new interpreter could break an old interpreter.

bin/civi-download-tools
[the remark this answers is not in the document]
totten: Fair question. My first inclination was to keep it (out of concern that we don't know if/how `nodesource` will change in the future; and we don't know how people will edit/copy/paste this fragment), but it's also valid to remove it like you did above (i.e. reducing duplication, improving execution time). TBH, the annualized ecosystem cost of either way is probably about the same. I just take path-of-least-resistance on this one.

bin/civi-download-tools
MikeyMJCO: Bionic does run a [newer version of bash](https://packages.ubuntu.com/search?suite=all&searchon=names&keywords=bash). Unfortunately I'm not sure of the root cause here but I can confirm that the new encoding/escaping works on Debian Jessie and Stretch, and Ubuntu 14.04+

totten: OK, nice testing then. Debian Jessie seems like a fair benchmark. Double-check, checked. ✅

MikeyMJCO: I didn't test with Precise Pangolin (I figure it's too old and it won't install BK anyway due to a nodejs issue)

totten: Hrm, well, I just checked the `bash` version that's bundled in OSX 10.12 (macOS sierra; first release Sept 2016). It comes with bash 3.2 and displays the message as:

```
WARNING:\ Failed\ to\ locate\ command\ "node".\ NodeJS\ \(http://nodejs.org/\)\ is\ required\ for\ development\ of\ CiviCRM\ v4.6+.
```

Situation TBD...

MikeyMJCO: Hmmm... Didn't think of bash on OSX.

MikeyMJCO: I had been thinking of a better way to handle that message within the function which doesn't seem to suffer from the same escaping issues.

totten: OK, quadruple-checked, and now I'm confused. On OSX with either bash 3.2 (via apple's stock build) or bash 4.4 (via nix's build), the message displays with funny escaping.

NOTE: To test how the message is displayed, I locally hack the `civi-download-tools` so that's looking for non-existent `nodezz` instead of `node`.  I also tweak the first `#!` line to make sure it doesn't grab the wrong interpreter.

MikeyMJCO: On Ubuntu 18.04.1 without the changes - it fatals the entire install process - which confused me!

totten: For posterity -- we had a long debugging session after this, starting circa

https://chat.civicrm.org/civicrm/pl/jcw3ibc6s7bfxbizkyjzxonmre

TLDR: the use of `pecl install` leads to a bizarre crash if you execute within a `curl ... | bash ... --full ...` context. Confusingly, it works fine if you execute locally, and tracing this required us to follow a lot of red-herrings. The resolution is to skip that and provide a warning instead.

totten:
👍 This now passes for me on a Bionic VM in the `curl "file://$PWD/civi-download-tools" | bash -s -- --full` test scenario.

PR #433, opened 2018-09-30
Debian Fixes

MikeyMJCO:
See issue #315 

Issues reported there are fixed in this PR. This PR also tweaks the Debian Stretch installation to remove a duplicate `apt-get update` and print the correct release name.

### DIFF
--- a/bin/civi-download-tools
+++ b/bin/civi-download-tools
@@ -486,7 +486,7 @@ function do_system_jessie() {
     echo "Recommended packages: $PACKAGES"
     echo ""
     if cvutil_confirm "Run automated installation? [Y/n] " y y; then
-      sudo apt-get update
+      wget -O- https://deb.nodesource.com/setup_8.x | sudo -E bash -
       sudo apt-get -y install $PACKAGES
       sudo php5enmod mcrypt
       sudo php5enmod imap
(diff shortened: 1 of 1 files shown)


totten:
Initially, `civi-download-tools --full` was targetting NodeJS circa v5(or 4? or 6?) because that would make it easier to run the Shoreditch build scripts. A few relevant happenings:

* In `civi-download-tools`, `do_system_stretch()` switched to 8.x circa Sep 2017
* In Shoreditch, the `README.md` was [updated](https://github.com/civicrm/org.civicrm.shoreditch/commit/571d58cbe89d7385013e00a36d1a55b05d10b697#diff-04c6e90faac2675aa89e2176d2eec7d8R14) about two weeks ago recommend 8.x.
* Test infra now tracks `bknix` -- in which the  the all three profiles (`min`/`max`/`dfl`) are currently using NodeJS 6.x.

I'd be interested in determining whether 8.x works with core's protractor and current Shoreditch; if so, we should bump up bknix max as well.

MikeyMJCO:
So I've run and tested with core on 8.x on Stretch and Jessie and on Ubuntu 14.04, 16.04 and 18.04.

MikeyMJCO:
More extensive testing should probably be performed beyond the installation where nodejs is used - however we're already at 8.x for Ubuntu 18.04 and Debian Stretch.

MikeyMJCO:
> I'd be interested in determining whether 8.x **works with core's protractor and current Shoreditch**; if so, we should bump up bknix max as well.


**Protractor:** Works fine.


**Current Shoreditch:** Cloning latest Shoreditch and `npm I` works fine - no errors and builds a working Shoreditch theme.

PR #434, opened 2018-10-03
PCHR-4158: Clean up civihr buildkit configs and point to master

AkA84:
This PR get rids of all the old civihr build configs and aliases, leaving only a generic `civihr` config and alias. Additionally, the config is now pointing to `master` instead of `staging`, as we're dropping the latter branch


AkA84:
@davialexandre please review when you have time

PR #435, opened 2018-10-04
Better PHP MySQL warning message

MikeyMJCO:
Follow up from discussion in #339 - if this PR is merged 339 should be closed.


eileenmcnaughton:
I suspect this is too dated to matter now - but seems safe & it does make it all clear

PR #437, opened 2018-10-11
Added info about build types to help files

hoegrammer:


src/help/create.hlp
```diff
@@ -2,12 +2,12 @@
...
 
 [33mArguments:[0m
-[32m  <build-name>[0m         The name of the sub-directory to build
+[32m  <build-name>[0m         String identifier for the site, which also becomes the subdirectory name unless otherwise specified. You can use one of the available site-types here (see below). If you don't, you must specify the site-type as an option.
 [32m  <ms-id>[0m              The unique id for a site within a multisite build
 
 [33mOptions:[0m
 [32m  -h, --help[0m           Displays help for a command
-[32m  --type <type>[0m        The name of the download/install scripts. (If omitted, assume <build-name>) [Optional]
+[32m  --type <type>[0m        The type of system to install. Must be one of the available types - see src/civibuild.aliases.sh (If omitted, assume <build-name>) [Optional]
```
totten: @hoegrammer Overall, good change.

A couple suggestions:

* I think it's good to show an example value. ("...The type of system to install. (Ex: drupal-demo)...")
* For reference of the available types, perhaps we could point to either [app/config/](https://github.com/civicrm/civicrm-buildkit/tree/master/app/config) (which has the authoritative list) or to https://docs.civicrm.org/dev/en/latest/tools/civibuild/#build-types (which is more prose-y).
* Generally, it might good for all the help screens to have a link at the bottom to the dev guide for `civibuild`?

hoegrammer: Done

totten:
Thanks, @hoegrammer ! This looks good to merge.

PR #440, opened 2018-12-07
Allow Drupal Clean installs to get all l10n files

seamuslee001:


seamuslee001:
This is exact port of what happens in drupal-demo install

totten:
Did a test build and confirmed that the localization files were downloaded and operational. :+1:

PR #441, opened 2018-12-14
Update to latest version of npm packages to fix vulerabilities

seamuslee001:
ping @totten @MikeyMJCO

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,12 +9,12 @@
     "url": "git://github.com/civicrm/civicrm-buildkit.git"
   },
   "devDependencies": {
-    "bower": "^1.8.3",
+    "bower": "^1.8.8",
     "civicrm-cv": "^0.1.2",
     "grunt-cli": "^1.2.0",
     "jasmine-core": "~3.3.0",
-    "jshint": "^2.5.11",
-    "karma": "~3.1.1",
+    "jshint": "^2.10.1",
+    "karma": "~4.0.0",
     "karma-chrome-launcher": "~2.2.0",
     "karma-jasmine": "~2.0.1",
     "karma-junit-reporter": "~1.2.0",
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "civicrm-cv": "^0.1.2",
     "grunt-cli": "^1.2.0",
     "jasmine-core": "~3.3.0",
-    "jshint": "^2.10.1",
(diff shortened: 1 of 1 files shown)


seamuslee001:
paging @totten

agh1:
I just tried this out and it resolves the problems @alifrumin was having [documented here](https://chat.civicrm.org/civicrm/pl/459n5tsa1pdwdfst8d4ifcnb3y)

totten:
Eh, this seems to require upgrading NodeJS? I spun up the [temporary worker](https://github.com/civicrm/civicrm-infra/blob/master/continuous-integration/worker-gen-3-gcloud.md) with this patch and built a clean site (`civibuild create dc --type drupal-clean --civi-ver 5.12`). Running `karma` on `bknix-min` now produces:

```
/home/jenkins/bknix-min/civicrm-buildkit/node_modules/karma/lib/server.js:118
  async start () {
        ^^^^^

SyntaxError: Unexpected identifier
    at createScript (vm.js:56:10)
    at Object.runInThisContext (vm.js:97:10)
    at Module._compile (module.js:549:28)
    at Object.Module._extensions..js (module.js:586:10)
    at Module.load (module.js:494:32)
    at tryModuleLoad (module.js:453:12)
    at Function.Module._load (module.js:445:3)
    at Module.require (module.js:504:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/home/jenkins/bknix-min/civicrm-buildkit/node_modules/karma/lib/cli.js:8:16)
```

The `bknix-min` profile has NodeJS 6.x. If I switch to NodeJS 8.x, then it seems to run.

The main reason we've used an older NodeJS was because Shoreditch was pretty particular at one point. But the [README](https://github.com/civicrm/org.civicrm.shoreditch) now points to NodeJS 8.x, so we're probably in fair shape to switch (obviously with proviso's about doing deployment and PSAs)?

seamuslee001:
yeh i guess so and also node v6 EOL is pretty dam soon as well @totten https://nodejs.org/en/about/releases/

agh1:
The node version may be the original issue: @alifrumin and I were both running NodeJS 8.x.

I just opened https://github.com/civicrm/civicrm-dev-docs/pull/594 to update the recommended version in the dev docs.

totten:
Good idea @agh1

PR #443, opened 2019-02-06
Cleanup composer.lock to reflect composer.json

eileenmcnaughton:
Over time things have been removed from composer.json (or never committted???) such that running composer update results in a very different json.lock

This commit are the results of running composer update. I made 2 composer.lock changes

1. I also increased min php version to 5.5 - this is crazy conservative as we no longer support 5.5 but....it should help us get a sensible packages set
2. I added  pear/console_table: 1.1.6 - this is newly important as drush
looks for it (& not the older 1.1.5 per the old lock) & if it can't find it tries to get it from http://pear.php.net/package/Console_Table
 - except it decommissioned

With the old lock file the semver prevents the console table package

Trace at https://integration.wikimedia.org/ci/job/wikimedia-fundraising-civicrm/7878/consoleFull

### DIFF
--- a/bin/dcg
+++ b/bin/dcg
@@ -1,0 +1,1 @@
+../vendor/chi-teck/drupal-code-generator/bin/dcg
--- a/bin/psysh
+++ b/bin/psysh
@@ -1,0 +1,1 @@
+../vendor/psy/psysh/bin/psysh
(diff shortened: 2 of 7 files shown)


eileenmcnaughton:
stale - closing

PR #444, opened 2019-02-12
Bump minimum version for upgrade tests to 4.2.9

colemanw:
In line with core bump - https://github.com/civicrm/civicrm-core/pull/13580

### DIFF
--- a/bin/civi-test-run
+++ b/bin/civi-test-run
@@ -192,7 +192,7 @@ fi
 for TESTTYPE in $SUITES ; do
   case "$TESTTYPE" in
     karma)  task_karma ;;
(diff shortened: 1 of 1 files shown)


colemanw:
@totten Is this the correct file to patch for Jenkins CI testing? civicrm/civicrm-core#13580 will break PR tests and the core matrix.

totten:
Yeah, that seems OK. Strictly speaking, the `@4.3..` expression will not match any snapshots from 4.2.x, but `4.2.9-multilingual_af_bg_en.mysql.bz2` is still explicitly in the list, so there is some coverage of upgrading from 4.2.9.

Aside: IMHO, we're due for generating some new snapshots... e.g. to represent various 5.x releases and probably add a newer multilingual variant.

PR #446, opened 2019-02-28
do not activate civicrm-admin-utilities on install

kcristiano:
https://lab.civicrm.org/dev/core/issues/765
Signed-off-by: Kevin Cristiano <kcristiano@kcristiano.com>


kcristiano:
@colemanw this will fix the issue on wpmaster.demo.civicrm.org while still having then plugin available for local development

colemanw:
Admin-utils are nice and I agree we should keep shipping buildkit with them, but I agree disabling them by default will reduce unexpected behavior.

PR #447, opened 2019-03-19
Remove recommendation to install mcrypt

eileenmcnaughton:
We are no longer recommending this in core since it's as likely to create problems as solve them - ie
adding mcrypt to a functional php7.2 site will result in problems with passwords.

We don't have a great fix for encryption but we should stop making this recommendation as we
are making a recommendation that we are not confident will be helpful and may cause harm. Better to make no recommendation than a bad one

### DIFF
--- a/bin/civi-download-tools
+++ b/bin/civi-download-tools
@@ -750,7 +750,7 @@ check_php_ext hash recommended
 check_php_ext imap recommended
 check_php_ext intl recommended
 check_php_ext mbstring recommended
-check_php_ext mcrypt recommended "WARNING: Failed to locate the PHP \"mcrypt\" extension. Please install this manually to avoid issues with SMTP passwords or older versions of CiviCRM."
+
 if php -r 'exit(version_compare(PHP_VERSION, 7.0, "<") ? 0 : 1);' ; then
   check_php_ext mysql recommended  ## In case you try to install older Civi versions (<= 4.7.11)
 fi


eileenmcnaughton:
@totten I think we should merge this - our current defacto position is that we don't encourage new installs to have mcrypt as they will be creating passwords that will have decryption problems, and there is no solution in progress so less instances of the problem is better. 

We removed the checks from core on the same basis

JoeMurray:
A few links on the general topic: mcrypt is no longer part of php in 7.2 but can be added via PECL: https://www.techrepublic.com/article/how-to-install-mcrypt-for-php-7-2/

I think going forward openssl is a good replacement that we could recommend. Here's one relatively seamless conversion based on a particular algorithm: https://stackoverflow.com/questions/9993909/replace-mcrypt-with-openssl OpenSSL is a very widely supported standard so I don't think it would be difficult for Civi hosting despite the heterogeneous platforms we need to support. 

As I recall there was debate about the security implications of having passwords for accessing services stored as plain text unencrypted in DB. I still tend to think this is a better practice, so I'd be reluctant to remove encryption of passwords even if it is only in 1 place.

eileenmcnaughton:
@JoeMurray sure - but we are removing a recommendation for something we won't be recommending here. We shouldn't be recommending something until we know what it is.  Especially since we can cause problems by recommending the wrong thing

JoeMurray:
+1 to merge for buildkit

PR #448, opened 2019-03-30
Fix up composer and ensure that only a php_codsniffer on the 2.x bran…

seamuslee001:
…ch is downloaded otherwise phpcs breaks

This fixes 2 issues 1. it seems composer.lock is no longer up to date with composer.json and 2. at present depending on your system composer update may cause the 3.x branch of codesniffer to be installed which breaks the phpcs this fixes it


totten:
The php codesniffer constraint makes sense. :+1:

The `composer.lock` doesn't work on PHP 5.6:

Screen Shot 2019-03-29 at 7 35 32 PM

However, I grabbed the older revision and did a narrower `composer update squizlabs/php_codesniffer` -- and that does seem to work. If you're OK with that, then go ahead and merge.

seamuslee001:
yeh that is fine with me @totten

PR #449, opened 2019-04-10
Upgrade coder to latest version

seamuslee001:
paging @totten

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,8 +18,8 @@
     "php": ">=5.3.3",
     "totten/php-symbol-diff": "dev-master#54f869ca68a3cd75f3386f8490870369733d2c23",
     "civicrm/upgrade-test": "dev-master#b93207fd28ac90426e9c95d572068f2e907b08e2",
-    "drupal/coder": "dev-8.x-2.x-civi#a801890a82d52e23a83c5d820270b6e228843d79",
-    "squizlabs/php_codesniffer": ">=2.7 <3.0"
+    "drupal/coder": "dev-8.x-2.x-civi#a1fab2860d916519bae5574b73067bb1c749b133",
+    "squizlabs/php_codesniffer": ">=2.7 <4.0"
   },
   "repositories": [
     {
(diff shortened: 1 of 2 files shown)


totten:
Force-pushed update so that `composer.lock` pulls in versions that work on PHP 5.6+.

@seamuslee001 It looks like 5.13 compat is in pretty good shape. :+1: At this point, the only hitch to activating is that `5.12` is still around and doing occasional PRs. Ideas:

* Defer deployment a couple weeks
* Deploy, but change the logic of the PR test job. Currently, it exits quickly if there's a style problem. We could mitigate by either:
    * If the branch is 5.12, then skip civilint entirely (As a policy matter, changes should only be going to 5.12 if they passed on 5.13. And, even if there is a small syntax mess-up, it won't matter next month.)
    * If the branch is 5.12, then run civilint... but don't exit on error.

```bash
#### Current snippet which executes civilint in CiviCRM-Core-PR
$GUARD pushd "$BLDDIR/sites/all/modules/civicrm"
  if git diff --name-only "origin/$ghprbTargetBranch.." | $GUARD civilint --checkstyle "$CHECKSTYLEDIR" - ; then
    echo "Style passed"
  else
    echo "Style error"
    exit 1
  fi
$GUARD popd
```

seamuslee001:
hmm I would probably be on the no exit but only on the 5.12 branch. My concern with leaving it some weeks is we may diverge from code standards that we have just brought the code up to

totten:
I'm down with that. Updated `CiviCRM-Core-PR`.

```bash
$GUARD pushd "$BLDDIR/sites/all/modules/civicrm"
  if git diff --name-only "origin/$ghprbTargetBranch.." | $GUARD civilint --checkstyle "$CHECKSTYLEDIR" - ; then
    echo "Style passed"
  else
    echo "Style error"
    ## TODO: Restore simpler logic after April 2019
    # exit 1
    [ "$ghprbTargetBranch" = "5.12" ] || exit 1
  fi
$GUARD popd
```

PR #454, opened 2019-06-01
Include PHPUnit6 in download

seamuslee001:


eileenmcnaughton:
seems like it's only additional - we can remove phpunit4 later....

PR #458, opened 2019-06-03
Use CiviNative TAP printer if we have it otherwise fall back to tap

seamuslee001:
This should work without https://github.com/civicrm/civicrm-core/pull/14421 but if 14421 is merged will use that printer instead of PHPUnit's tap

### DIFF
--- a/bin/phpunit-each
+++ b/bin/phpunit-each
@@ -11,7 +11,16 @@ $hasError = 0;
 
 chdir("$srcDir/tools");
 foreach ($args as $test) {
-  $cmd = sprintf("./scripts/phpunit --tap --log-junit %s %s %s",
+  if (file_exists("../Civi/Test/TAP.php")) {
+    $printCMD = "--printer '\Civi\Test\TAP'";
+  }
+  elseif (class_exists('PHPUnit_Util_Log_TAP')) {
+    $printCMD = "--tap";
+  }
+  else {
+    $printCMD = "--debug";
+  }
+  $cmd = sprintf("./scripts/phpunit {$printCMD} --log-junit %s %s %s",
     escapeshellarg("$xmlDir/$test.xml"),
     $options,
     escapeshellarg($test)


bin/phpunit-each
```diff
@@ -11,7 +11,16 @@ $hasError = 0;
 
 chdir("$srcDir/tools");
 foreach ($args as $test) {
-  $cmd = sprintf("./scripts/phpunit --tap --log-junit %s %s %s",
+  if (file_exists("../Civi/Test/TAP.php")) {
+    $printCMD = "--printer '\Civi\Test\TAP'";
+  }
+  elseif (class_exists('PHPUnit_Util_Log_TAP')) {
```
totten: I'm confused - this is standalone PHP script, and it hasn't done any setup for autoloading -- so how would this check every return true?

Here's an alternative way to think about substituting `--tap` with `--printer \Civi\Test\TAP`: https://gist.github.com/totten/acfb8845dca00caef1420b78961139b5 . With the gist, the polyfill mechanism for TAP support goes entirely in `civicrm-core.git` (ie both the code `Civi/Test/TAP.php` and the conditional to activate it), so we don't have to worry about these edge-cases.

bin/phpunit-each
```diff
@@ -11,7 +11,16 @@ $hasError = 0;
...
+    $printCMD = "--tap";
+  }
+  else {
+    $printCMD = "--debug";
+  }
```
totten: I think there's a basic question "*How important is TAP format*?" There are a few different reasons one might use `--tap`:

1. It's handy for a person skimming the output (esp when trying to gauge progress and/or to investigate PHP fatals).
2. It's a formal spec which can be integrated/scripted with other tools. In this case, it's part of a technical contract.

If `#1` completely captures our usage, then we have a lot of flexibility. It's acceptable to have some mid-tier scripts which switch around between output formats (e.g. `--tap` vs `--debug`). It may even be preferable to change from TAP format to a custom format (borrowing good bits from `--tap` and `--debug`). OTOH, if it's `#2`, then other scripts would expect the output to be strictly TAP -- so switching to anything else (even if it's a fallback like `else {$printCMD = '--debug';}`) would break the contract.

bin/phpunit-each
seamuslee001: Yeh ok can do that. Are you able to tac that onto my core TAP PR @totten?

seamuslee001:
For mine we have been using tap on our Jenkins boxes for ages and switching phpunit versions shouldn’t necessarily change the output we see in the console log

bin/phpunit-each
totten: Sure, will do momentarily.

PR #460, opened 2019-06-21
civibuild - Standardize on "web/" subfolders. Prep for D8+VDR. 

totten:
Overview
---------

This patch-set revises the `{drupal,backdrop,wp}-{demo,clean,empty}` builds to  use a `web/` folder. For example, in a `dmaster` (`drupal-demo`) site, the path to the CiviCRM code changes:

* Old: `/home/myuser/buildkit/build/dmaster/sites/all/modules/civicrm`
* New: `/home/myuser/buildkit/build/dmaster/web/sites/all/modules/civicrm`

This addresses a test-infrastructure problem with rolling out support for D8 as an equal CMS, and it has the side benefit of aligning the various build-types to a closer technical standard.

I'd like this PR to be open/WIP for a bit so that there's room for folks to feedback on issues with this transition.

Transition Notes (Typical)
-------------------------

The change is *not* retroactive to pre-existing builds. Thus, if you have a working setup with the old structure, you can still use it. Even `civibuild restore` and `civibuild reinstall` should continue working. However, if you make a new build (`civibuild create`) or do a total-rebuild/forced-create (`civibuild create --force`), then that it will use the new `web/` folder. 

When/if you create a new build, you may need to change some customizations/integrations/references. For example, if you use  PHPStorm + PHPUnit + civibuild, then this configuration will change:
 
Screen Shot 2019-06-20 at 4 07 59 PM-highlight

The change is targeted at the build-types which are most used in civicrm.org's CI -- i.e. `{drupal,backdrop,wp}-{demo,clean,empty}` and `site-list`. Other custom build-types are not affected.

Transition Notes (test*.civicrm.org / VDR)
-------------------------

There is an exception to the previous note -- if the httpd is configured using Apache VirtualDocumentRoot (VDR), then the transition (*without-`web/` => with-`web/`*) needs to be done all at once. This configuration is not common on personal workstations - it mostly affects the civicrm.org CI servers.

When deploying to `test*.civicrm.org`, we will probably flush out any recent CI builds.

Background / Goal
-------------------

* The test infrastructure generates+hosts a large number of builds (with lifespans commonly up to ~1 week). The most common is `drupal-clean` (used for PR-testing with `civicrm-core.git`), but the other CMS builds (such as `backdrop-clean` and `wp-demo`) are regularly used as well (e.g. for PR-testing other repos and for building out QA matrices).
* We generally want to add representative D8 builds into the test systems -- so that the Civi-D8 integration gets the same level of continuous-testing as the other CMSs.
* The test systems use Apache `VirtualDocumentRoot` (VDR) to host the large number of builds concurrently (while keeping resource-usage and system-restarts down). VDR allows you to define virtual-hosts with a simple file-naming pattern. (*For our purposes, VDR only accepts **one** file-naming pattern.*)
* However, there is a mismatch in the file-naming convention for document roots (`build/<BLDNAME>` vs `build/<BLDNAME>/web`). This makes deployment more complicated. This conflict already exists today - but only affects the less-visible micro-services, so it's not very symptomatic. When we roll out D8 as an equal CMS, the mismatch would become more problematic. Unless we make some change so that the builds are better aligned...

Before
-------
* The build types in `app/config/*` mostly abide one of two file conventions:
    * *`build/<BLDNAME>` as HTTP document root*: This includes the most commonly used build-types of the form `{backdrop,drupal,wp}-{clean,demo,empty}`.
    * *`build/<BLDNAME>/web` as HTTP document root*: This currently includes build-types for some existing/internal microservices (e.g. `messages`, `dist`, `distmgr`). (*However, as we expand D8 to include environments that are more representative of Civi-D8 early-adopters, it'll need this.*)
* Buildkit uses amp v0.4.11

After
-----
* The build types in `app/config/*` mostly abide one file convention, `build/<BLDNAME>/web`.
* Buildkit uses amp v0.5.0 with a more nuanced [DB naming convention](https://github.com/amp-cli/amp/pull/54).

Technical Details
--------------------

I wrestled a fair bit with how to approach this issue.  You can see some of the alternatives and trade-offs in [Mattermost logs](https://chat.civicrm.org/civicrm/pl/cyft6z4fz3n9fprpddwzji3iee).

The main strength of this approach is that it leaves us with a simpler and more consistent end-state than the other approaches (KISS!) - while keeping the general performance/resource-variables the same.

My main fear is that updating CI standards will break local-workstations. To mitigate that possibility, I've been following a fairly aggressive QA matrix to ensure that the transition-behavior is somewhat understandable. Specifically:

* Pre-existing builds continue to work, regardless of whether they had a `web/` subfolder. You can safely issue `civibuild reinstall` or `civibuild restore`
* New builds (`civibuild create`) and force-rebuilds (`civibuild create --force`) do have a `web/` subfolder. If you blow-away the codebase, then you'll get the new one.
* To determine changes between builds, I'm using this one-off script [cfgsnap](https://gist.github.com/totten/2d2c0e18f098e547012592eacb2990f2).

```
                        ORIG    PATCH   UP-RESTORE      UP-REINSTALL    UP-CREATE       UP-REINSTALL-FL UP-CREATE-FL
WASFLAT drupal-empty    X       X       X               X               X               -               -
WASFLAT backdrop-empty  X       X       X               X               X               -               -
WASFLAT wp-empty        X       X       X               X               X               -               -

WASFLAT site-list       X       X       X               X               X               X               X
WASDEEP messages        X       -       X               X               X               -               -
WASDEEP dist            X       -       X               X               X               -               -
WASDEEP distmgr         X       -       X               X               X               -               -

WASFLAT drupal-clean    X       X       X               X               X               X               X
WASFLAT backdrop-clean  X       X       X               X               X               -               -

WASFLAT backdrop-demo   X       X       X               X               X               -               -
WASFLAT drupal-demo     X       X       X               X               X               X               X
WASFLAT wp-demo         X       X       X               X               X               X               X

Task Definitions:
 * ORIG: Generate a site with old scripts. Save a snapshot of all code+metadata.
 * PATCH: Write a draft patch to follow new directory structure by default
 * UP-RESTORE: Checkout updated/patched scripts. Do a `civibuild restore`. For WASFLAT|WASDEEP, the configs (esp vhost+db names) should look unchanged.
 * UP-REINSTALL: Checkout updated/patched scripts. Do a `civibuild reinstall`. For WASFLAT|WASDEEP, the configs (esp vhost+db names) should look unchanged.
 * UP-CREATE: Checkout updated/patched scripts. Do a `civibuild create --force`. For WASFLAT, the configs and code should change.
 * UP-REINSTALL-FL: Like UP-REINSTALL, but also use the webapp in a browser (login/click around)
 * UP-CREATE-FL: Like UP-CREATE, but also use the webapp in a browser (login/click around)

Status Definitions:
* "X": Pass, OK
* "-": Skip, not testing. This item is either unnecessary or reasonably inferred by a combination of other checks.
* "?": Unknown, not yet tested
```

Todo
------

Before merging this, we will also make an alternate branch available in case one needs a stable reference to the previous build-scripts.


michaelmcandrew:
Thinking about compatibility with https://github.com/michaelmcandrew/civicrm-buildkit-docker...

I had a quick look at the files that you have changed.

The [Dockerfile](https://github.com/michaelmcandrew/civicrm-buildkit-docker/blob/master/civicrm/Dockerfile#L122) references a [custom apache24 template](https://github.com/michaelmcandrew/civicrm-buildkit-docker/blob/master/civicrm/apache24-vhost.php) but the only difference between that file and the [amp default template](https://github.com/amp-cli/amp/blob/master/src/Amp/views/apache24-vhost.php) (which I see has not changed recently) is that I added a couple of lines to export logs outside of the container.

Was there anything else specific that you were worried about?

totten:
Thanks, @michaelmcandrew. I just pinged you as general precaution/alert. The custom apache24 relies appropriately on the `$root`. 👍  The fact that civicrm-buildkit-docker uses [httpd_type: apache24](https://github.com/michaelmcandrew/civicrm-buildkit-docker/blob/dd194d9079625deab25b18d49dcce22d633db09f/civicrm/amp.services.yml#L8) should mean it's generally flexible about variations in root path.  🤞

michaelmcandrew:
Sounds good

JoeMurray:
Not (really) trying to bikeshed, but I think a more standard subdir name than /web/ would be /htdocs/
+1 for initiative overall.

MikeyMJCO:
I'm not sure I agree with /htdocs/ as a standard tbh.

I think /web/ works - but in reality there isn't much of a standard.

Nginx=www
Apache=www or htdocs
IIS=inetpub
Litespeed=html

Not that any of those is truly important - this is a buildkit internal "thing" and really we're standardising for our benefit as users of buildkit rather than some external standard.

As long as we document it - it's good imo.

totten:
Yeah, I agree, @MikeyMJCO @JoeMurray, that the industry has several conventions and the choice is a little arbitrary. As a matter of happenstance, `web/` is most convenient - because it's already used in several `app/config/*` types, and it's used in [drupal-composer/drupal-project](https://www.drupal.org/docs/develop/using-composer/using-composer-to-install-drupal-and-manage-dependencies) (which has been circulating among Civi-D8 early adopters).

(*Etymological trivia: the happenstance probably is not an accident... `web/` was used as the default in Symfony v1.x, v2.x, v3.x, and I'd wager this influenced our micro-services as well as `drupal-project.git`*)

JoeMurray:
Good to know /web/ is already used somewhere in CiviCRM and thus provides a good rationale, and the conformity with Symfony is plausible and another good rationale. If there were no good such reasons, I was proposing the one we have used internally for 15+ years based on Apache, which historically dominated the web server market, and still slightly outranks nginx in market share. But it's all bike shedding.

Cheers,

totten:
OK, merging this so I can proceed with deployment to test systems. When that's done, I'll send another PSA to `civicrm-dev@`.

PR #467, opened 2019-08-20
civi-test-run - Use phpunit5 for both Civi{Drupal,WP} suites

totten:
These two test-suites run differently from the `civicrm-core` test suite, so they don't track the phpunit updates. So... update them...

### DIFF
--- a/bin/civi-test-run
+++ b/bin/civi-test-run
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -e
+PHPUNIT=phpunit5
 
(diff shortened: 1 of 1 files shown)


seamuslee001:
why not phpunit6?

totten:
Because that won't work? 🙃 

Just filed https://lab.civicrm.org/dev/core/issues/1194 - I put the "Sprint task" label on it because it seemed like a decent starter task for someone new to testing.

PR #468, opened 2019-08-21
civi-download-tools - Add drush-backdrop plugin

totten:
Before
------

On Backdrop builds, `drush` cannot be used - because the drush-backdrop plugin is missing.

After
-----

On Backdrop builds, `drush` can be used - because the plugin is present.

Comments
--------

This has come up a few ways - e.g.

* Test/demo sites don't have a test/demo user - https://lab.civicrm.org/infra/ops/issues/906
* When QA'ing patches for civicrm-backdrop's copy of `civicrm.drush.inc` - e.g. https://github.com/civicrm/civicrm-backdrop/pull/98

### DIFF
--- a/app/config/backdrop-clean/install.sh
+++ b/app/config/backdrop-clean/install.sh
@@ -33,4 +33,11 @@ civicrm_install
 
 pushd "$CMS_ROOT" >> /dev/null
   php "$SITE_CONFIG_DIR/module-enable.php" civicrm
+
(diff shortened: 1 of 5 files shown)


totten:
I've also tossed in a partial fix for infra/ops#906 - i.e. creating the `demo` user (but not creating the role/permissions). I'll post more discussion in the issue about that - but IMHO it's not a reason to block this bit.

PR #469, opened 2019-08-22
Migrate PHAR downloads from bash to composer-downloads-plugin

totten:
Before
------

There are ~16 PHAR/ZIP files downloaded via bash script in `civi-download-tools`

After
-----

Those same ~16 PHAR/ZIP files are downlaoded via `composer-downloads-plugin`.

Comments
--------

* Still allows multiple versions of different tools (i.e. phpunit5.phar + phpunit6.phar).
* The notation is more pithy/readable (less far-flung/less boiler-plate)
* When switching back/forth between revisions of buildkit with different versions of each tool, you'd get the benefit composer's cache

I've done some `r-run` on one laptop, but this is a bit of change, so we
should do some `r-run` in other contexts.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,7 @@
     "totten/php-symbol-diff": "dev-master#54f869ca68a3cd75f3386f8490870369733d2c23",
     "civicrm/upgrade-test": "dev-master#b93207fd28ac90426e9c95d572068f2e907b08e2",
     "drupal/coder": "dev-8.x-2.x-civi#a1fab2860d916519bae5574b73067bb1c749b133",
+    "civicrm/composer-downloads-plugin": "~2.0",
     "squizlabs/php_codesniffer": ">=2.7 <4.0"
   },
   "repositories": [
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -30,5 +31,25 @@
       "type": "git",
       "url": "https://github.com/totten/paratest.git"
     }
-  ]
+  ],
+  "extra": {
+    "downloads": {
+      "amp": {"url": "https://download.civicrm.org/amp/amp.phar-2019-06-25-dc96b3da", "path": "bin/amp", "type": "phar"},
+      "box": {"url": "https://github.com/box-project/box2/releases/download/2.7.5/box-2.7.5.phar", "path": "bin/box", "type": "phar"},
+      "civici": {"url": "https://download.civicrm.org/civici/civici-0.1.3.phar", "path": "bin/civici", "type": "phar"},
```
seamuslee001: @totten just a small thing i notice that the composer is saying these are all essentially dev-master but that might be confusing when there is an actual version in the url? is there a version param we can specify also?

seamuslee001: That feels sensible to me right

PR #471, opened 2019-08-28
Add environment variable for customizing the "app/config" types

jaapjansma:
With this PR it is possible to declare a configuration option `EXTRA_SITE_CONFIG_DIR` this directory could contain custom site types (without overriding existing ones).

### DIFF
--- a/app/civibuild.conf.tmpl
+++ b/app/civibuild.conf.tmpl
@@ -35,11 +35,11 @@
 # APACHE_VHOST_ALIAS=1
 
 #############################################################################
-## Example: an extra site configuration directory. You can put your custop site
+## Example: an extra site configuration directory. You can put your custom site
 ## types in this directory.
 ## No trailing slash.
 
-# EXTRA_SITE_CONFIG_DIR = "$HOME/extra_site_configs"
+# CIVIBUILD_PATH="$HOME/extra_site_configs:$CIVIBUILD_PATH"
 
 #############################################################################
 ## Example: Place all builds in /srv/www.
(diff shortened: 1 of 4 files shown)


jaapjansma:
Ah I see now the PR #391 from @michaelmcandrew which is similair

totten:
@jaapjansma Oh, nice! I think this approach is an improvement over #391:

* You don't have type `--type <foo>` frequently. The customization is persistent (`civibuild.conf`).
* But if you want to make the customization momentarily (akin to passing `--type`), you can just pass the env-var at runtime
    ```
    env EXTRA_SITE_CONFIG_DIR=...foo... civibuild create mybuild`
    ```
* If you want the change to be persistent across all builds, then you can put it in `civibuild.conf`.

I tested it out locally, and it worked. 😃 

A few nitpicks about the variable name:

* `EXTRA_SITE_CONFIG_DIR` looks to be named after `SITE_CONFIG_DIR` (which is closely related). But `EXTRA_` implies a sort of interchangeable content; in reality, the content has a different grain. (`SITE_CONFIG_DIR` is *a child path` which contains *the specific configuration used by a given site*. `EXTRA_SITE_CONFIG_DIR` is a place to run search.)
* In Unix variable naming, if you're setting a place to search, it's usually named `xxxPATH` (e.g. `PATH`, `MANPATH`, `LD_LIBRARY_PATH`, `CPATH`, `CLASSPATH`).

I pushed up a change to rename it to `CIVIBUILD_PATH`. If that looks good to you, then I'd say let's merge.

(*Will wait a couple days before merging in case there's feedback.*)

jaapjansma:
I am happy with your improvement.

totten:
Cool. Proposed update for dev docs: https://github.com/civicrm/civicrm-dev-docs/pull/666

PR #475, opened 2019-09-17
For 5.17 and 5.18 versions bind to using 4.5.2 of apiv4 otherwise dow…

seamuslee001:
…nload from the extensions directory. Eventually we will want to stop this on master/5.19 and up but that can come later

This pegs the version downloaded to be a specific tag version of apiv4 because 5.17 will only download 4.4.x versions of api4 

Eventually we will want a conditional on master and other branches until we get rid of 5.17/5.18 etc

### DIFF
--- a/app/config/backdrop-demo/download.sh
+++ b/app/config/backdrop-demo/download.sh
@@ -20,14 +20,7 @@ pushd "$WEB_ROOT/web/modules" >> /dev/null
   git clone ${CACHE_DIR}/civicrm/civicrm-core.git      -b "$CIVI_VERSION"     civicrm
   git clone ${CACHE_DIR}/civicrm/civicrm-backdrop.git  -b "1.x-$CIVI_VERSION" civicrm/backdrop
   git clone ${CACHE_DIR}/civicrm/civicrm-packages.git  -b "$CIVI_VERSION"     civicrm/packages
-  case "$CIVI_VERSION" in
-    5.17|5.18)
-     git clone "${CACHE_DIR}/civicrm/api4.git"                                -b "4.5.2"             civicrm/ext/api4
-     ;;
-  *)
-     EXTCIVIVER=$( php -r '$x=simplexml_load_file("civicrm/xml/version.xml"); echo $x->version_no;' )
-     cv dl -b "@https://civicrm.org/extdir/ver=$EXTCIVIVER|cms=Drupal|status=|ready=/org.civicrm.api4.xml" --to="$WEB_ROOT/web/modules/civicrm/ext/api4" --dev
-  esac
+  api4_download_conditional civicrm                                           civicrm/ext/api4
   git_set_hooks civicrm-drupal      civicrm/backdrop   "../tools/scripts/git"
   git_set_hooks civicrm-core        civicrm            "tools/scripts/git"
   git_set_hooks civicrm-packages    civicrm/packages   "../tools/scripts/git"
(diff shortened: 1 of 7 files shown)


app/config/backdrop-demo/download.sh
```diff
@@ -20,8 +20,14 @@ pushd "$WEB_ROOT/web/modules" >> /dev/null
...
-  git clone ${CACHE_DIR}/civicrm/api4.git              -b "master"            civicrm/ext/api4
-
+  case "$CIVI_VERSION" in
+    5.17|5.18)
+     git clone "${CACHE_DIR}/civicrm/api4.git"                                -b "4.5.2"             civicrm/ext/api4
+     ;;
+  *)
+     EXTCIVIVER=$( php -r '$x=simplexml_load_file("civicrm/xml/version.xml"); echo $x->version_no;' )
+     cv dl -b "@https://civicrm.org/extdir/ver=$EXTCIVIVER|cms=Drupal|status=|ready=/org.civicrm.api4.xml" --to="$WEB_ROOT/web/modules/civicrm/ext/api4" --dev
+  esac
```
totten: @seamuslee001 A few thoughts...

* I suppose we're not likely to remember to update this when/if there's a new tag of `4.5.x`.  There are git branches for previous versions (`4.4`, `4.3`) -- so I guess it makes sense to add one for `4.5`. (Done.)
* The `info.xml` for `api4@4.5` says that it requires `<ver>5.18</ver>`; and here were going for `4.5` on either `5.17|5.18`. This begs the question of whether `5.17` is actually compatible with `4.5`. I'll do a little test...
* Supposing that `4.5` is published... maybe the `cv dl -b` command is the only one we need? Don't need the `git clone`?

totten: * For purposes of merging `api4` into `civicrm-core` (https://github.com/civicrm/org.civicrm.api4/issues/198), we actually want to *not* download `api4` going forward. The basic rationale for checking out `api4@4.5` is (paradoxically) to allow `api_v3_AllTests` to continue passing on `5.17` and `5.18`.

totten:
@seamuslee001 I pushed an update to:

* Consolidate the special-case logic for different version handling of api4 download
* Change that logic - skip the download completely if the folder `$civicrm_root/Civi/Api4` exists; and if it doesn't exist, then use the conditional version.

I'm not certain what the right pairings are for `civicrm-core@5.14,5.15,5.16` + `api4@??`; but I don't think it matters much since we're not testing them now.

seamuslee001:
@totten the only cases maybe for 5.13 testing in Core-Matrix but hopefully in a couple of releases time we will fade out needing to download anything

seamuslee001:
@totten your changes look good to me maybe just add in some handling for 5.13 then we are happy

totten:
Before pushing, I did a trial run with api4 downloaded in a few couple ways:

```
## Create builds with api4 in core - mixing `master` plus patch
for T in backdrop-demo drupal-clean drupal-demo wp-case wp-demo; do 
  civibuild download "del$T" --type "$T" --force --civi-ver master --patch https://github.com/civicrm/civicrm-core/pull/15309
done

## Create builds api4 ext
for T in backdrop-demo drupal-clean drupal-demo wp-case wp-demo; do
  civibuild download "del$T" --type "$T" --force --civi-ver 5.17
done
```

totten:
> @totten your changes look good to me maybe just add in some handling for 5.13 then we are happy

Yeah, alright, I dug into the history of `civicrm-core`, and I think that `3e20c1acb397d6dfe` looks like the turning-point wherein `api4` became a bidirectional dependency. That commit shows up in `5.16.0`, so that should mean:

* `ver <=5.15`  - Don't need `api4` with core
* `5.16 <= ver <= ~5.19.alpha1` - Do need `api4` with core, roughly v4.5
* `ver >= ~5.19.alpha1` - Planned turning point at which `api4` is part of core

seamuslee001:
:+1: looks good LGTM

totten:
Yup, let's give it a go

PR #477, opened 2019-10-09
Condititionally enable apiv4 extension on demos if it is appropriate

seamuslee001:
This should fix the erorrs we are having with the demo sites

ping @MikeyMJCO

### DIFF
--- a/src/civibuild.lib.sh
+++ b/src/civibuild.lib.sh
@@ -617,7 +617,9 @@ function civicrm_apply_demo_defaults() {
       ));
 EOPHP
   fi
-  cv en --ignore-missing api4
+  if cv ev 'exit(version_compare(CRM_Utils_System::version(), "5.19", "<") ?0:1);' ; then 
+    cv en --ignore-missing api4
+  fi
 }
 
 ###############################################################################


MikeyMJCO:
Reviewed and tested locally ok.

src/civibuild.lib.sh
```diff
@@ -617,7 +617,9 @@ function civicrm_apply_demo_defaults() {
       ));
 EOPHP
   fi
-  cv en --ignore-missing api4
+  if cv ev 'exit(version_compare(CRM_Utils_System::version(), "5.19", "<") ?0:1);' ; then 
+    cv en --ignore-missing api4
+  fi
```
totten: Are you sure we need these extra conditionals? The `api4` extension is supposed to be omitted on new builds:

https://github.com/civicrm/civicrm-buildkit/pull/475/commits/c2d8c68f2e1d5a4e57f9cdda15d1d55f567189ab

If the demo system is throwing up an error error, then that means that the above conditional isn't working as expected.

I think the problem is just that `www-demo` was out-of-date. 

However, the activation conditional is also useful for situations like this -- you keep a build on your local workstation for a long time and periodically switch among different revisions of Civi.

PR #478, opened 2019-10-19
Add export UI to demo sites

eileenmcnaughton:
I think we should definitely put this on demo sites.

I would also argue for Contact Layout Editor & Afform - I realise afform is still moving but there is enough there that I think it's better for people to have it - esp since dev sites are demo sites

### DIFF
--- a/app/config/drupal-demo/download.sh
+++ b/app/config/drupal-demo/download.sh
@@ -28,7 +28,7 @@ pushd "$WEB_ROOT/web"
     git clone "${CACHE_DIR}/civicrm/org.civicoop.civirules.git"              -b "$RULES_VERSION"     civicrm/tools/extensions/org.civicoop.civirules
     git clone "${CACHE_DIR}/TechToThePeople/civisualize.git"                 -b "master"             civicrm/tools/extensions/civisualize
     git clone "${CACHE_DIR}/dlobo/org.civicrm.module.cividiscount.git"       -b "$DISC_VERSION"      civicrm/tools/extensions/cividiscount
-    git clone "${CACHE_DIR}/colemanw/exportui.git                    "       -b "master"             civicrm/tools/extensions/exportui
+    git clone "${CACHE_DIR}/colemanw/exportui.git"       -b "master"             civicrm/tools/extensions/exportui
     extract-url --cache-ttl 172800 civicrm=http://download.civicrm.org/civicrm-l10n-core/archives/civicrm-l10n-daily.tar.gz
     ## or https://raw.github.com/civicrm/l10n/master/po/fr_CA/civicrm.mo => civicrm/l10n/fr_CA/LC_MESSAGES/
   popd
--- a/app/config/drupal-demo/download.sh
+++ b/app/config/drupal-demo/download.sh
@@ -28,7 +28,7 @@ pushd "$WEB_ROOT/web"
     git clone "${CACHE_DIR}/civicrm/org.civicoop.civirules.git"              -b "$RULES_VERSION"     civicrm/tools/extensions/org.civicoop.civirules
     git clone "${CACHE_DIR}/TechToThePeople/civisualize.git"                 -b "master"             civicrm/tools/extensions/civisualize
     git clone "${CACHE_DIR}/dlobo/org.civicrm.module.cividiscount.git"       -b "$DISC_VERSION"      civicrm/tools/extensions/cividiscount
-
+    git clone "${CACHE_DIR}/colemanw/exportui.git                    "       -b "master"             civicrm/tools/extensions/exportui
     extract-url --cache-ttl 172800 civicrm=http://download.civicrm.org/civicrm-l10n-core/archives/civicrm-l10n-daily.tar.gz
     ## or https://raw.github.com/civicrm/l10n/master/po/fr_CA/civicrm.mo => civicrm/l10n/fr_CA/LC_MESSAGES/
   popd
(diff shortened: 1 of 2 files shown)


eileenmcnaughton:
@colemanw

seamuslee001:
@eileenmcnaughton don't you also need to create a download step as well a la https://github.com/civicrm/civicrm-buildkit/blob/master/app/config/drupal-demo/download.sh#L30

eileenmcnaughton:
@seamuslee001 thanks - I added that too

kcristiano:
I agree that Contact Layout Editor & Afform should be added.  I also have a small bit of cleanup on wp-demo as I need to remove a plugin that won't be needed after WP 5.3 is released.  If there is consesus on the new extensions, I can do the cleanup PR and add the new extensions.

eileenmcnaughton:
@colemanw can we merge this?

colemanw:
@eileenmcnaughton IMO we should just merge the extension into core. Assuming it's had sufficient testing on your end?

eileenmcnaughton:
@colemanw We have used it & it works but I also think there is a bit of work getting it into core since it requires ripping stuff out too so I think it will just stall getting wider exposure if we don't merge this

colemanw:
I'd be happy to submit a PR for the ripping out and merging the extension into core.

eileenmcnaughton:
@colemanw ok go for it

eileenmcnaughton:
I should note we only use it with shoreditch - I still vote for merging this now & it can come out again when you get around to the rip out & it gets reviewed & merged

eileenmcnaughton:
(that approach is also much closer to the agreed approach & not having an interim period in buildkit probably added to the snaffus with KAM)

app/config/drupal-demo/download.sh
```diff
@@ -28,7 +28,7 @@ pushd "$WEB_ROOT/web"
     git clone "${CACHE_DIR}/civicrm/org.civicoop.civirules.git"              -b "$RULES_VERSION"     civicrm/tools/extensions/org.civicoop.civirules
     git clone "${CACHE_DIR}/TechToThePeople/civisualize.git"                 -b "master"             civicrm/tools/extensions/civisualize
     git clone "${CACHE_DIR}/dlobo/org.civicrm.module.cividiscount.git"       -b "$DISC_VERSION"      civicrm/tools/extensions/cividiscount
-
+    git clone "${CACHE_DIR}/colemanw/exportui.git                    "       -b "master"             civicrm/tools/extensions/exportui
```
colemanw: There appears to be extra whitespace at the end of this string.

colemanw:
Ok. This seems mergeable except for the whitespace issue.

eileenmcnaughton:
OK - I removed it

eileenmcnaughton:
cool - next up CLE & afform :-)

eileenmcnaughton:
@kcristiano I think you had a pr to do this for wordpress & joomla

kcristiano:
I did, let me see if it needs a rebase for WP

app/config/drupal-demo/download.sh
```diff
@@ -28,7 +28,7 @@ pushd "$WEB_ROOT/web"
     git clone "${CACHE_DIR}/civicrm/org.civicoop.civirules.git"              -b "$RULES_VERSION"     civicrm/tools/extensions/org.civicoop.civirules
     git clone "${CACHE_DIR}/TechToThePeople/civisualize.git"                 -b "master"             civicrm/tools/extensions/civisualize
     git clone "${CACHE_DIR}/dlobo/org.civicrm.module.cividiscount.git"       -b "$DISC_VERSION"      civicrm/tools/extensions/cividiscount
-
+    git clone "${CACHE_DIR}/colemanw/exportui.git"       -b "master"             civicrm/tools/extensions/exportui
```
totten: Note that this was failing on new builds because the folder `${CACHE_DIR}/colemanw/exportui.git` didn't exist. For any reference into `CACHE_DIR`, there should be an antecedent call to `git_cache_setup`. 

Calls to `git_cache_setup` are conventionally placed either:

* In `app/config/caches.sh`... if the repo is used frequently.
* In `app/config/FOO/download.sh`... if the repo is only used once.

From context, it looks like you're aiming to include in each `*-demo` build-type, so  ce9f42a70995784ff95850df16ddd8be0554790b updates `app/config/caches.sh`.

app/config/drupal-demo/install.sh
```diff
@@ -110,6 +110,7 @@ EOPERM
   drush -y cvapi extension.install key=org.civicrm.angularprofiles debug=1
 
   drush -y cvapi extension.install key=org.civicrm.volunteer debug=1
+  drush -y cvapi extension.install key=exportUI debug=1
```
totten: Also worth noting that https://github.com/colemanw/exportui/blob/master/info.xml lists the extension's name as `exportui` not `exportUI`, so this line also fails. Fixed in 5e0b0c8b2fec6aee6d70438eb3834a634a5b16ef.

totten:
Note: If you do a `civibuild reinstall <name>` on an existing `drupal-demo` build, then you'll likely get:

```
+++ drush -y cvapi extension.install key=exportui debug=1
Unknown extension: exportui                                                                                                                                                                 [error]
Array
(
    [is_error] => 1
    [error_message] => Unknown extension: exportui
    [xdebug] => Array
        (
            [peakMemory] => 37176104
            [memory] => 37114864
            [timeIndex] => 3.7522921562195
        )

)
```

You can resolve with:

```
cd sites/all/modules/civicrm/tools/extensions/
git clone https://github.com/colemanw/exportui
```

eileenmcnaughton:
Maybe we should get the contactlayout editor into the install list now so people only have to deal with it once for both

eileenmcnaughton:
Oh & afform

PR #481, opened 2019-10-25
Change days before stale and messaging

MikeyMJCO:
Drops the value of `days-before-stale` from 60 to 30 and introduces more useful messaging.

If we like this we should cascade across the other Repos with actions turned on for this.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -13,7 +13,8 @@ jobs:
     - uses: actions/stale@v1
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


totten:
I like the expanded wording. :+1:

Personally, for something like a stale/timeout, I have a strong lean towards longer periods (like 60d or 90d).

Are there other civicrm repos that have this turned on? I thought we were still in a trialing phase with just one.

eileenmcnaughton:
lol - it's stale :-)

I'm gonna merge this as 'better' - but I note that we really want to switch to the version that removes stale again when activity happens

PR #485, opened 2019-11-05
Use standard function to get civiversion and re-write if statments following civi-version retrival change

seamuslee001:
…llowing civi-version retrival change

@totten this should work perfectly now tested both the clean and demo versions locally and looks correct to me


totten:
Cool 👍  Looks sensible. Merging+deploying to `www-demo`...

PR #488, opened 2019-11-25
wp-demo, wp-case - Initialize wpLoadPhp

totten:
This is a follow-up to https://lab.civicrm.org/dev/core/issues/1412 (especially https://github.com/civicrm/civicrm-core/pull/15922). The issue yielded a new E2E test.

For the `wp-cli` use-case, the test implicitly relies on the setting `wpLoadPhp`, which is usually initialized on the first browser-based page-view of the admin UI. However, in scripted E2E, we don't start out with a browser-based page-view.

Aside: It's a bit frustrating that we still don't have a good unified place for install logic. I'd vote for switching to `civicrm-setup` by default, and then you can initialize WP-specific path settings [here](https://github.com/civicrm/civicrm-setup/blob/01019db/plugins/init/WordPress.civi-setup.php#L60-L64).

### DIFF
--- a/app/config/wp-case/install.sh
+++ b/app/config/wp-case/install.sh
@@ -42,6 +42,7 @@ wp theme install twentythirteen --activate
 wp eval '$home = get_page_by_title("Welcome to CiviCRM with WordPress"); update_option("page_on_front", $home->ID); update_option("show_on_front", "page");'
 
 wp plugin activate civicrm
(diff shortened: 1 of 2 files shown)


kcristiano:
@totten Patch looks good and an `r-run` on both `wp-demo` and `wp-case` works locally.   Value set after install and all cli utils are available.

totten:
Nice. Thanks @kcristiano

totten:
Deployed to `test-{1,2,3}`

PR #489, opened 2019-12-16
Allow for backdrop installs to run the upgrade tests

seamuslee001:
This PR depends on the changes made to the upgrade-test repo here https://github.com/civicrm/civicrm-upgrade-test/pull/3/files, I have tested this on my local dev and seems to work fine


totten:
Good idea. The code reads right, and I expect your local testing was representative. :+1:

totten:
By the by... at some point, we may want to switch this to `cv upgrade:db` so that it's possible to run upgrade tests on all CMSs. (But that'll be a bit of work to re-validate that the test routine emits accurate signals.)

seamuslee001:
@totten note that for it to be successful it needs this change as well https://github.com/civicrm/civicrm-upgrade-test/pull/3/files

PR #490, opened 2019-12-25
Switch drupal, wordpress, backdrop phpunit test runner to be phpunit6…

seamuslee001:
… for civiversions above 5.21

Now that PRs for https://lab.civicrm.org/dev/core/issues/1194 have been merged as detailed in https://lab.civicrm.org/dev/core/issues/1194#note_29301 i figure we should do this

### DIFF
--- a/bin/civi-test-run
+++ b/bin/civi-test-run
@@ -1,6 +1,14 @@
 #!/bin/bash
 set -e
-PHPUNIT=phpunit5
+
+function getPhpunitVer() {
+  CIVIVER=$(getCiviVer)
+  if [ $(php -r "echo version_compare('$CIVIVER', '5.21', '>=');") ]; then
+    echo "phpunit6 --printer \Civi\Test\TAP"
+  else
+    echo 'phpunit5 --tap'
+  fi
+}
 
 #################################################
 ## Helpers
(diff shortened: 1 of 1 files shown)


bin/civi-test-run
```diff
@@ -1,6 +1,14 @@
 #!/bin/bash
 set -e
-PHPUNIT=phpunit5
+
+function getPhpunitVer() {
+  CIVIVER=$(getCiviVer)
+  if [ $(php -r "echo version_compare('$CIVIVER', '5.21', '>=');") ]; then
```
totten: On the good - on a local build with 5.22, that updated command was generally running :+1:

To check this detail (`5.21`), I did a little spot-checking - from https://github.com/civicrm/civicrm-wordpress/pull/175/commits , determine the commit id 541cfd457de9032f8f1ed568fb1863f6417ddf93 and check which branches have the commit.

```
[bknix-max:~/bknix/build/wpmaster/web/wp-content/plugins/civicrm] git branch --contains 541cfd457de9032f8f1ed568fb1863f6417ddf93
  5.22
  master
```

So doesn't this mean it should be >= 5.22?

seamuslee001: ah yeh oops

seamuslee001:
fixed @totten

totten:
Yup, confirmed that fixed running on 5.21.

PR #491, opened 2020-01-02
Work around Drupal issue #2978575 by manually applying patch to fix M…

seamuslee001:
…ySQL 8 compliance

This works around the Drupal issue here https://www.drupal.org/project/drupal/issues/2978575 where D7 installation fails on MySQL 8 because of the sql_modes and an issue with the system table

ping @totten @monishdeb @JoeMurray


seamuslee001:
Note this affects Backdrop as well so have created a function for backdrop and done a similar fix for it

totten:
Cool. This seems generally sensible. :+1:

As a style matter, for functions in `civibuild.lib.sh`, it's helpful to include an assertion about referenced variables. I threw that it in and caught some pre-existing bugs around the Backdrop downloads and CMS_VERSION.

I tested this (in combination with #492 and with some regression-fixes following #478). It was able to build various permutations of `{drupal,backdrop}-{empty,clean,demo}` and `bknix-{min,edge}` - for the permutations that I tried, it appeared to produce a working site (where you could login via web UI or SQL CLI). There may still be issues within certain configs (e.g. extra PHP warnings), but this does make the build better. :)

PR #497, opened 2020-01-24
Download the l10n daily file for wp-demo and backdrop install types

seamuslee001:


totten:
Thanks @seamuslee001. Change looks pretty sensible. I did an `r-run` on `wp-demo` (`civibuild create delme-1-1 --type wp-demo --civi-ver 5.22`) and confirmed that:

* It generates localized data files.
* The new `tests/phpunit/E2E/Core/LocalizedDataTest.php` passes.

PR #505, opened 2020-02-14
Fix WP E2E Unit tests runs

seamuslee001:
This fixed the E2E test failure on the path url test locally for me however i would be curious if @kcristiano  or @christianwach  had any ideas as to why this is needed in the first place

### DIFF
--- a/app/config/wp-case/install.sh
+++ b/app/config/wp-case/install.sh
@@ -48,6 +48,7 @@ wp plugin activate civicrm-demo-wp
 echo '{"enable_components":["CiviMail","CiviReport","CiviCase"]}' | cv api setting.create --in=json
 civicrm_apply_demo_defaults
 cv ev 'if(is_callable(array("CRM_Core_BAO_CMSUser","synchronize"))){CRM_Core_BAO_CMSUser::synchronize(FALSE);}else{CRM_Utils_System::synchronizeUsers();}'
+wp eval 'civicrm_activate();'
 
 ## Install Shoreditch and CiviCase
 cv en shoreditch styleguide civicase
--- a/app/config/wp-demo/install.sh
+++ b/app/config/wp-demo/install.sh
@@ -65,7 +65,7 @@ wp plugin install gutenberg-ramp --activate
 
 civicrm_apply_demo_defaults
 cv ev 'if(is_callable(array("CRM_Core_BAO_CMSUser","synchronize"))){CRM_Core_BAO_CMSUser::synchronize(FALSE);}else{CRM_Utils_System::synchronizeUsers();}'
-
+wp eval 'civicrm_activate();'
 wp role create civicrm_admin 'CiviCRM Administrator'
 wp cap add civicrm_admin \
   read \


kcristiano:
@seamuslee001 

Added PR to buildkit.  Built wpcvmaster

Without loading the site, ran `./tools/scripts/phpunit 'E2E_AllTests'`

First Notable item:
```
PHP Notice:  Undefined index: ADMIN_USER in /srv/www/buildkit/wpcvmaster/web/wp-content/plugins/civicrm/civicrm/tests/phpunit/E2E/Cache/CacheTestCase.php on line 24
```

```
FAILURES!
Tests: 1426, Assertions: 1656, Failures: 8, Skipped: 424.
```

First failure is the permissions error.  

` E2E\Core\PathUrlTest::testSystemRouter`

``` 
<div class="crm-section crm-error-message">You do not have permission to access this page.</div>\n
```

Load Website, load CiviCRM

Now same error:

```
PHP Notice:  Undefined index: ADMIN_USER in /srv/www/buildkit/wpcvmaster/web/wp-content/plugins/civicrm/civicrm/tests/phpunit/E2E/Cache/CacheTestCase.php on line 24

```

But results are 'better':

```
FAILURES!
Tests: 1426, Assertions: 1656, Failures: 7, Skipped: 424.
```
```
1) E2E\Core\PrevNextTest::testFillSql
Failed asserting that 0 matches expected 25.

/srv/www/buildkit/wpcvmaster/web/wp-content/plugins/civicrm/civicrm/tests/phpunit/E2E/Core/PrevNextTest.php:62
/home/kcristiano/projects/buildkit/extern/phpunit6/phpunit6.phar:570
```

The same assertion fails each time.

On a hunch I reverted https://github.com/civicrm/civicrm-core/pull/16284/

That removes the first error on permissions as I suspected.  As CleanerURLs needs that initial load to work properly, 

we still get 
````
FAILURES!
Tests: 1426, Assertions: 1656, Failures: 7, Skipped: 424.
```
After loading the web site and reruning the tests I get 
```
FAILURES!
Tests: 1426, Assertions: 1656, Failures: 7, Skipped: 424.
```

We can revert https://github.com/civicrm/civicrm-core/pull/16284/ if that will get E2E working.  

Once we correct the load issue that @christianwach mentioned https://chat.civicrm.org/civicrm/pl/eght1swdhbdaxyshc1hat539go  we should be able to get both E2E and clean urls working together.

Thoughts?

seamuslee001:
hmm that is odd what type of buildkit install is `wpcvmaster` ?

kcristiano:
yes wpcvmaster.   

```
civibuild create  wpcvmaster --type wp-demo --url http://wpcvmaster.test --web-root /srv/www/buildkit/wpcvmaster --admin-pass admin
```

seamuslee001:
@kcristiano so what i notice is that the anon role isn't created when we enabled only when i login and this function gets triggered https://github.com/civicrm/civicrm-wordpress/blob/master/includes/civicrm.users.php#L248

kcristiano:
@seamuslee001 Yes, that is an issue on automated builds.

I added:
```
wp eval '$c=[civi_wp()->users->set_wp_user_capabilities()];if (is_callable($c)) $c();'
```
at ln 154 and then all roles will be created before the UT run.   

I now get a different failure on `E2E\Core\PathUrlTest` as all the HTML is output on the home page.

Could the issue be that starting with 5.23 WP has CleanURLs by default?   And the E2E test fails on that?

If I disable CLEANURLs I get 7 failures on `E2E\Core\PrevNextTest` but no issues on `E2E\Core\PathUrlTest`

kcristiano:
@seamuslee001 I created https://github.com/civicrm/civicrm-buildkit/pull/509  as the changes here are throwing errors on build, calling `civicrm_activate` won't get the anonymous user created.

totten:
So if #509 is merged, should we also merge this?

seamuslee001:
@totten I’m no longer certain if it working right. I think there needs to register some of the page redirects in the wp integration but haven’t yet worked out exactly what.

kcristiano:
I don't think this should be merged, activating civicrm seems to only throw and error and not move us forward towards the solution.

christianwach:
Let me have a crack at refactoring the activation process next week and then we can revisit this.

kcristiano:
@christianwach That would be great.

I created https://github.com/civicrm/civicrm-buildkit/pull/510   before I saw this.   This forces the basepage creation via wp-cli.

totten:
So if I'm understanding correctly, the issue here is that one needs setup-tasks to run near the end of installation (i.e. after loading the SQL schema but before the user works with Civi UI). These setup-tasks are currently deferred until first administrative page-view. Specific tasks mentioned in Kevin's PRs:

```php
civi_wp()->users->set_wp_user_capabilities();
civi_wp()->basepage->create_wp_basepage();
```

The current implementation of the "on-first-login" mechanism is something that I don't fully grok, but loosely it involves the `civicrm_activate()` / `CiviCRM_For_WordPress->activate()` and the event `civicrm_activation`. (Hopefully I'm following.)

When using an automated install routine (like in E2E testing; but also conceivable for other cases), there is no "first administrative page-view" to key off. A few techniques for bridging that gap are:

* #505 tries to fire off all deferred setup-tasks with `wp eval 'civicrm_activate();'`
* #509 + #510 don't fire the event (`civicrm_activation`) at all - instead, they fire the tasks one-by-one.
* In the same spirit of #505, I think a tighter implementation might be 
`wp eval 'civi_wp(); do_action("civicrm_activation");'` (*Based on some printf debugging, i think this actually manages to call more of the setup tasks. And I don't see any session errors stemming from this command.*)

I'd suggest taking a critical view of the "on first login" technique. We've also used the "on first login" technique in `civicrm-core` (e.g. to call `System.flush` on first login), but TBH I think it was a mistake (architecturally speaking). The reason for that technique was that I was afraid to do anything with the `install/*.php` code - it's hard to read, it's hard to test, and it's not used reliably by all installers. However, using this work-around doesn't solve that underlying problem, and it created an arbitrary line between setup-tasks that are in the "installer" and ones which are "first-run". It counter-intuitively means the "installer" does not finish the installation. It further encourages ossification of the code in `install/*.php` (*Do anything else but fix the installer code!*)

`civicrm-setup` aims to address that root problem by refactoring `install/*.php`. The refactored version:

* Organizes tasks into more [bite-sized chunks](https://github.com/civicrm/civicrm-setup/blob/master/plugins)
* Allows any task to participate at any point during the install routine (using events and priorities)
* Allows you to inspect the list of tasks by browsing the file-system, running a command (`cv core:install --debug-event`), or with code.
* Allows adding/removing tasks from different sources (e.g. `civicrm-setup.git` and `civicrm-wordpress.git`) ([doc](https://github.com/civicrm/civicrm-setup/blob/master/docs/plugins.md)).

(Related example: `install/index.php` has some D7 setup-tasks akin to those WP tasks.  When refactoring for `civicrm-setup`, they landed in [installDatabase/FlushDrupal.civi-setup.php](https://github.com/civicrm/civicrm-setup/blob/master/plugins/installDatabase/FlushDrupal.civi-setup.php). If you drop more `*.civi-setup.php` files in there, they'll get used.)

I'd really like to start [switching to civicrm-setup in earnest](https://lab.civicrm.org/dev/core/issues/1615) circa 5.24/5.25/5.26.

Of course, for purposes of 5.23, we need something. Personally, I think it could make sense to either:

* Do the #509+#510 approach (itemize the steps individually in `install.sh`)
* Provide some short-term function like `civicrm_activate()` in `civicrm-wordpress.git` whose job is to fire the needful setup-tasks.

kcristiano:
@totten I think you are spot on.

As #505 caused errors for me, I went with a step by step approach.   Kind of whack-a-mole.    I'll try your suggested call to activate - I want to see how it works.  I thought step by step was better as we already activated.  let's see.

This should be moot by 5.24 as @christianwach has said he wants to look at the activate and init of the civ-wp process and streamline.  

What really bugs me is why did it not fail before now.

totten:
> (@kcristiano) What really bugs me is why did it not fail before now.

There might be an element of coincidence - e.g. 5.23 happens to include both (a)  updates to WP integration and (b) updates to E2E testing. In particular, the `PathUrlTest` is a new test-case. We didn't have coverage like that before.

kcristiano:
@totten I agree that the best approach is to refactor the activation so it does not rely on a first administrative page view.   I also want to get us towards using `civicrm-setup` as the default.

For now, I tested using:

`wp eval 'civi_wp(); do_action("civicrm_activation");'`

in place of 

```
wp eval '$c=[civi_wp()->users->set_wp_user_capabilities()];if (is_callable($c)) $c();'
wp eval '$c=[civi_wp()->basepage->create_wp_basepage()];if (is_callable($c)) $c();'
```
as my way felt whack-a-mole.

When I then ran the follwoing just after install of master via Buildkit on wp:

```
wp eval 'civicrm_initialize(); echo CRM_Utils_System::url("civicrm/mailing/subscribe", "reset=1", TRUE, NULL, FALSE, TRUE);'
cv ev 'echo CRM_Utils_System::url("civicrm/mailing/subscribe", "reset=1", TRUE, NULL, FALSE, TRUE);' --level=cms-full  
cv ev 'echo CRM_Utils_System::url("civicrm/mailing/subscribe", "reset=1", TRUE, NULL, FALSE, TRUE);' --level=full
```
All of them returned URLs without the base page.

The subsequent tests `./tools/scripts/phpunit 'E2E_AllTests'`  failed the URL test.

I do think we need to merge #510 in the short term to have the tests pass and then we can revert once the refactoring is complete.

PR #506, opened 2020-02-15
Fix d8 composer install on master branch now that the cache integrati…

seamuslee001:
…on tests issue has been fixed


totten:
@seamuslee001 Policy updates look good. I ran a build locally and spot-checked a few patches to confirm it was working. :+1:

One change in the implementation - to set the `extra.enable-patching` option, we don't need to use a PHP script. `composer config` can handle that part.

Merging.

PR #508, opened 2020-02-18
Add in test function to execute Core Extension unit tests

seamuslee001:


eileenmcnaughton:
ohh thank you - can we merge it & see how Jenkins looks? I'm inclined to make it scan & find other folders that are added later - but that doesn't have to be this round

bin/civi-test-run
```diff
@@ -256,6 +271,7 @@ for TESTTYPE in $SUITES ; do
     phpunit-drupal) task_phpunit_drupal ;;
     phpunit-wordpress) task_phpunit_wordpress ;;
     phpunit-backdrop) task_phpunit_backdrop ;;
+    phpunit-creditnotes) task_phpunit_core_extension sequentialcreditnotes Financial ;;
```
totten: I'm not sure I like the file structure of "civicrm-core:CoreExtensions/Financial/sequentialcreditnotes". But I'm not criticizing this PR for that. However, if that changed, then you'd have to edit both the callers and content of the function.

One of these would feel like a cleaner separation of concerns to me:

```
## Give both full-key and full-path
task_phpunit_extension "sequentialcreditnotes" "$CIVI_CORE/CoreExtensions/Financial/sequentialcreditnotes"

## Just give a path. 
task_phpunit_extension "$CIVI_CORE/CoreExtensions/Financial/sequentialcreditnotes"
# note: The function can figure out the key using `cd $path; civix info:get -x @key`)

## Just give a key. 
task_phpunit_extension "sequentialcreditnotes"
# note: The function can figure out the path using `cv path -x $key`)
```

totten:
@seamuslee001 So are you thinking to just add this as another item i https://test.civicrm.org/job/CiviCRM-Core-Matrix/ ?

My initial assumption was that the `civi-test-run ... all` (eg for https://test.civicrm.org/job/CiviCRM-Core-PR/ ) would  include the `sequentialcreditnotes`?

seamuslee001:
@totten either or but we still needed a way to trigger the test (I think I forgot to update the all suites thing btw)

eileenmcnaughton:
@totten @seamuslee001 yes - I think the tests in the CoreExtensions folder should be treated as part of core & run whenever core tests are run

bin/civi-test-run
eileenmcnaughton: I was rather hoping any phpunit.xml files under the CoreExtensions would be automagically picked up rather than us keeping adding them as we add new ones

seamuslee001: Sadly @eileenmcnaughton that is not the case even when we added in api_v4 we needed to add in another test suite function for it

totten:
OK. Agree with preference to do scanning, but I don't think we need to block this for now.

Added a little tidying so that the new suite shows up in the help screen and when using the `all` target.

PR #509, opened 2020-02-19
Create Anonymous user role via wp-cli for E2E test

kcristiano:
alternative to https://github.com/civicrm/civicrm-buildkit/pull/505

### DIFF
--- a/app/config/wp-case/install.sh
+++ b/app/config/wp-case/install.sh
@@ -106,3 +106,6 @@ wp cap add civicrm_admin \
   view_public_civimail_content
 
 wp user create "$DEMO_USER" "$DEMO_EMAIL" --role=civicrm_admin --user_pass="$DEMO_PASS"
+
+## Ceate anonymous user role
+wp eval '$c=[civi_wp()->users->set_wp_user_capabilities()];if (is_callable($c)) $c();'
--- a/app/config/wp-demo/install.sh
+++ b/app/config/wp-demo/install.sh
@@ -151,6 +151,8 @@ wp cap add civicrm_admin \
   view_public_civimail_content
 
 wp user create "$DEMO_USER" "$DEMO_EMAIL" --role=civicrm_admin --user_pass="$DEMO_PASS"
+## Ceate anonymous user role
+wp eval '$c=[civi_wp()->users->set_wp_user_capabilities()];if (is_callable($c)) $c();'
 
 cv en angularprofiles volunteer
 


totten:
I've been experimenting a bit with this and #505. I can confirm locally that this combination produces passing E2E tests: apply #509 and revert https://github.com/civicrm/civicrm-core/pull/16284 (i.e. don't set `CIVICRM_CLEANURL` to `1`). I'm going to go ahead and merge #509 because it seems to be orthogonal  to the clean-url issue.

Turning to the clean-url issue... this is the first time I've tried using clean-urls on WP, and I'm a bit confused.

The E2E test is generating a URL with

```php
 \CRM_Utils_System::url('civicrm/mailing/subscribe', 'reset=1', TRUE, NULL, FALSE)
```

So I hacked in some code to check the environment and see what this URL looks like:

```php
  public function testSystemRouter() {
    print_r([
      'CIVICRM_CLEANURL' => \CRM_Utils_Constant::value('CIVICRM_CLEANURL') ? 'true' : 'false',
      'url' =>  \CRM_Utils_System::url('civicrm/mailing/subscribe', 'reset=1', TRUE, NULL, FALSE),
    ]);
```

Regardless of whether `civicrm.settings.php` has the new conditional, the debug output is always this:

```
    [CIVICRM_CLEANURL] => false
    [url] => http://wpmaster.bknix:8001/?page=CiviCRM&q=civicrm%2Fmailing%2Fsubscribe&reset=1
```

This leads me a few thought/questions (most unresolved; in no particular order):

1. Why is `CIVICRM_CLEANURL` false while executing the E2E test? The E2E tests should be using the same `civicrm.settings.php` as normal. (*Perhaps relevant that `tests/phpunit/CiviTest/bootstrap.php` calls `cv php:boot --level=settings`. For E2E testing, it might be better to use `--level=cms-full`, though obviously that wouldn't work for headless tests.*)
2. If `CIVICRM_CLEANURL` is true on WP, what is the URL supposed to look like? With the value being true, I guessed  and tried `http://wpmaster.bknix:8001/civicrm/mailing/subscribe?reset=1` - but that gives a 404.
3. There's a difference between *generating* URLs (clean/dirty) and *accepting* URLs (clean/dirty). IIRC, in D7, the `CIVICRM_CLEANURL` toggle only dictates what Civi *generates*.... it always *accepts* dirty URLs even if it *prefers to generate* clean URLs.

totten:
A couple informational bits - these don't resolve points above but may help with interpreting...

* For the specific route of `CRM_Utils_System::url('civicrm/mailing/subscribe',...)`, it appears that other links provide the `frontend=TRUE` option. (Doh!) https://github.com/civicrm/civicrm-core/pull/16607

* With regard to what the value of `CIVICRM_CLEANURL` is while executing E2E tests, these seem like interesting debugging bits:

```
## Make URL with cv's "--level=full"
$ cv ev 'echo CRM_Utils_System::url("civicrm/mailing/subscribe", "reset=1", TRUE, NULL, FALSE, TRUE);' --level=full
http://wpmaster.bknix:8001/?page=CiviCRM&q=civicrm%2Fmailing%2Fsubscribe&reset=1

## Make URL with cv's "--level=cms-full"
$ cv ev 'echo CRM_Utils_System::url("civicrm/mailing/subscribe", "reset=1", TRUE, NULL, FALSE, TRUE);' --level=cms-full
http://wpmaster.bknix:8001/mailing/subscribe/?reset=1

## Make URL with wp cli
$ wp eval 'civicrm_initialize(); echo CRM_Utils_System::url("civicrm/mailing/subscribe", "reset=1", TRUE, NULL, FALSE, TRUE);'
http://wpmaster.bknix:8001/mailing/subscribe/?reset=1
```

kcristiano:
@totten Thanks for the detailed analysis.

For CleanURLs to work in WP, we have to flkush WP's rewrite rules.  This is done in the plugin and on the first page load of CiviCRM.

If I build wp-demo with master, and after the build I go to the site and do not login, the shortcode URLS work.

The Ugly URLs do not.    This is due to the basepage not being ready. `civicrm`.

Once I log in Then the Ugly URLs work.   

I have to load CiviCRM to get CLEANURLs to not 404.    This is not an issue on a UI install as we load CiviCRM after install.    

Further with CLEANURLs set to false, I still get 404 errors on any page that uses the basepage if there has been no WP login by an admin.

In either case logging in allows all URLs to work.

This matches Tim's results above.    The issue is the basepage is missing from the URLs,

I have not found what we need to do for automated cli testing.   I'll keep digging.  I believe this is all related to the current way we create the basepage and  that we rely on the UI loads.   

I'll keep digging.  Thanks for the push in the right direction.

PR #510, opened 2020-02-21
Create basepage for E2E tests

kcristiano:
Currently the basepage is not created until the UI loads.   This loads it via wp-cli.


seamuslee001:
I tested it and it worked, however we may be better to wait for @christianwach  to try a refactoring of the plugin activation

totten:
We're criss-crossing threads a bit, but in the other thread Kevin followed-up and suggested merging anyway. Refactoring is good, but it seems unlikely to get into 5.23.

Since @seamuslee001 says this was working for him, I'm gonna merge.

Whenever we do have updates to the install mechanism, then we can remove/replace the two itemized tasks from #509 and #510.

PR #512, opened 2020-03-20
Gutenberg is here to stay

kcristiano:
It's improved to the point we should not disable on test sites.  Remove gutenberg-ramp which disables the block editor, install and activate classic editor which allows switching between the new block editor and the classic editor


kcristiano:
Reopening.   I do think we need to change the way we handle the block editor (aka Gutenberg) Early on it was better for development to disable, but now we need it on.

MikeyMJCO:
I agree it'd be nice to have this enabled as it would be on someone's Wordpress site if they just installed it.

kcristiano:
Thanks I agree.  We also need this PR before we can add the CiviCRM blocks @mecachisenros is working on for testing.

The benefit of using blocks is that it could be   adapted to any CMS using Gutenberg

MikeyMJCO:
Very cool - we're playing with it on D8.

PR #517, opened 2020-05-14
Github Workflow - Update the "stale" action to use v3

MikeyMJCO:
Adds support for removing the stale label on activity.


totten:
Sounds sensible. Let's give it a whirl.

PR #518, opened 2020-05-26
Update Drush version to be 8.3.3 to support MySQL8, php7.4

seamuslee001:
also ping @MikeyMJCO


MikeyMJCO:
Can't see an issue with this - tests okay on our BuildKit box.

totten:
Sweet :)

totten:
I may have been hasty -- reading their patch for mysql80 support, I suspect it broke <mysql57. But I guess we'll find out...

PR #519, opened 2020-06-01
Upgrade Karma and Protractor versions to their latest versions

seamuslee001:
@totten


totten:
Looks good.

I ran this locally and had a tangential issue (b/c [this line](https://github.com/totten/bknix/blob/master-loco/.loco/loco.yml#L47) in bknix) -- but this patch seems good. :+1:

PR #522, opened 2020-06-06
Add in a drupal9-clean build

seamuslee001:
ping @totten @MikeyMJCO


seamuslee001:
So this has been found to work for me if I apply the following PRs:

https://github.com/civicrm/civicrm-core/pull/17515
https://github.com/civicrm/civicrm-core/pull/17522
https://github.com/civicrm/civicrm-drupal-8/pull/40

And built custom versions of cv with this patch https://github.com/civicrm/cv/pull/75 and drush8 with this patch incorporated https://github.com/drush-ops/drush/pull/4431

seamuslee001:
So the only thing I think that is misssing for our purposes is for there to be a drush8 version that includes that Role PR apart from that this should work correctly on dev-master

seamuslee001:
@MikeyMJCO are you able to test this please?

demeritcowboy:
Tested on WSL. Didn't try anything fancy just checked the site loaded and the civi menus weren't all screwy.

At the beginning drush warned about not being able to find any supported drupal version but then installed 9.0.3 anyway. Otherwise ok.

seamuslee001:
Merging as per @demeritcowboy 's review

PR #527, opened 2020-06-25
Fix failing to copy upgrade test result xml following test server con…

seamuslee001:
…fig change

@totten I have deployed this for the DFL build on test-1 and running a manual test now

### DIFF
--- a/.loco/config/mysql-common/my.cnf.php
+++ b/.loco/config/mysql-common/my.cnf.php
@@ -37,6 +37,7 @@ innodb_file_per_table = 1
 <?php if (matchVer('/Ver 5.6/')) { ?>
 innodb_large_prefix = 1
 <?php } ?>
+<?php if (!matchVer('/Ver 8.0/')) { ?>innodb_file_format = Barracuda<?php } ?>
 
 <?php if (matchVer('/Ver 5.7/')) { ?>
 # https://expressionengine.com/blog/mysql-5.7-server-os-x-has-gone-away
--- a/bin/civi-test-run
+++ b/bin/civi-test-run
@@ -121,7 +121,7 @@ function task_upgrade() {
   fi
   echo "Found EXITCODES=\"$EXITCODES\""
   if [ -d "$PRJDIR/build/$BLDNAME/.civibuild" ]; then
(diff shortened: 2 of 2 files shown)


seamuslee001:
I'm going to merge this in so I can run this on test-1 where it is showing good promise

bin/civi-test-run
```diff
@@ -118,7 +120,11 @@ function task_upgrade() {
     EXITCODES="$EXITCODES upgrade-test"
   fi
   echo "Found EXITCODES=\"$EXITCODES\""
-  $GUARD cp "$PRJDIR/app/debug/$BLDNAME/civicrm-upgrade-test.xml" "$JUNITDIR/"
+  if [ -d "$PRJDIR/build/$BLDNAME/.civibuild" ]; then
```
totten, line 123: Yup, this looks sensible.

.loco/config/mysql-common/my.cnf.php
```diff
@@ -37,6 +37,7 @@ function matchVer($pat) {return (bool) preg_match($pat, ver());}
 <?php if (matchVer('/Ver 5.6/')) { ?>
 innodb_large_prefix = 1
 <?php } ?>
+<?php if (!matchVer('/Ver 8.0/')) { ?>innodb_file_format = Barracuda<?php } ?>
```
seamuslee001, line 40: @totten this was to fix some Min test failures I have no idea how to actually apply this so found the relevant files and manually fixed them on the server

totten, line 40: I like how much easier that is now. :)

Small revision: https://github.com/civicrm/civicrm-buildkit/pull/528

totten: Good q. Easiest ways to apply this:

* For a local environment (where one uses `loco run`), just add the `-f` / `--force` option (i.e. `loco run -f`).
* For CI environment (where there are several systemd services), use `nix/bin/reset-ci.sh`.

In both cases, it'll re-generate the concrete config files and start the services.

(There are also more fine-tuned options like `loco init -f mysql`.)

PR #539, opened 2020-08-11
Rpow config for WMF

eileenmcnaughton:
I'm looking to add rpow to our buildkit - but I can't figure out how to set the dsns

@totten @mepps I'll close this again soon as it's clearly incomplete - but wanting to try to get thoughts as to how to populate the DSNs in it

### DIFF
--- a/app/config/wmff/install.sh
+++ b/app/config/wmff/install.sh
@@ -35,6 +35,9 @@ CIVI_EXT_URL="${CMS_URL}/sites/default/civicrm/extensions"
 
 civicrm_install
 
+chmod 774 "${WEB_ROOT}/sites/default/civicrm/extensions/rpow/bin/harvey-dent"
+"${WEB_ROOT}/sites/default/civicrm/extensions/rpow/bin/harvey-dent" --root "${WEB_ROOT}/drupal"
+
 ###############################################################################
 ## Extra configuration
 pushd "$CMS_ROOT"


eileenmcnaughton:
@totten this relates to the one you just commented on - I'm really keen to pick your brains on it because I can't see my way through right now

totten:
@eileenmcnaughton Before trying to answer technically, let me ask about the deployment topology you're trying to get at. In the past, I used two topologies for different occasions:

1. Run one `mysqld` with two Civi databases (and run a slow/semi-manual sync between databases). This is handy if you want to inspect how it works during different parts of sync.
2. Run two `mysqld`'s, each with one Civi database (and run mysql's real-time(-ish) sync mechanism). This is more representative of runtime.

The setup and workflow are different for these topologies.

eileenmcnaughton:
@totten honestly I was just gonna cheat locally & have them both point back to the same database/ dsn - which is very cheaty - but it at least allows us to run the module in dev instances. Ideally we would have 2 users, one being read only & we would at least see any glaring issues.

I'm aiming for kinda minimal 'we can at least step through the code' - eg. to test the drush thing locally/ spot any really obvious breakage on new releases

If I could generate the rpow above so it has our dsns in it when I run civibuild.create that would be enough to meet that limited requirement

totten:
@eileenmcnaughton Oh, that's pretty clever. I like it.

Added a section to https://github.com/totten/rpow/blob/master/doc/install.md

Specifically: https://github.com/totten/rpow/blob/master/doc/install.md#using-civibuild-and-harvey-dent

eileenmcnaughton:
@totten thanks I'll work through that to see if I can get it to run on our ci civibuild!

eileenmcnaughton:
@mepps do you want to test this ? It worked for me &- but I did have some resource url issues - which I think were unrelated

 I can merge to this repo. This folder in the repo only affects WMF but I want to check the resource urls are not a regression

eileenmcnaughton:
I'm going to merge this & test on our CI as it only affects WMF

PR #542, opened 2020-08-14
Remove exportUI from install

eileenmcnaughton:
Cleanup - this is now in core

### DIFF
--- a/app/config/drupal-demo/install.sh
+++ b/app/config/drupal-demo/install.sh
@@ -110,7 +110,6 @@ EOPERM
   drush -y cvapi extension.install key=org.civicrm.angularprofiles debug=1
 
   drush -y cvapi extension.install key=org.civicrm.volunteer debug=1
-  drush -y cvapi extension.install key=exportui debug=1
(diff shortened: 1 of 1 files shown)


seamuslee001:
@eileenmcnaughton given that the civi-ver param means any CiviCRM version could be installed I would be reluctant to do this because these are templates and the Extension sub system should quietly ignore it if you ask for it to be installed

eileenmcnaughton:
@seamuslee001 well - it was only really added for the demo sites so people would see it. There is no particular need for it to be enabled

seamuslee001:
@eileenmcnaughton my point is that I would expect if you did `civibuild create 520-demo --type drupal-demo --civi-ver 5.20` that it should enable the extension but doing it like `civibuild create 527-demo --type drupal-demo--civi-ver 5.27` shouldn't cause a failure in the demo install process and that the export ui extension shouldn't be enabled because it is no longer needed but other than that and the CiviCRM versions the builds should be identical

eileenmcnaughton:
@seamuslee001 right - but I don't know it should - I mean - we really did only add it in to give it visibility for that period when it wasn't in core & we wanted some eyes on

PR #544, opened 2020-08-14
D8/D9 demo - turn on errors on screen

demeritcowboy:
In drupal 8/9 error logging to screen is completely off by default.

I'd argue it should be on for the "clean" versions too but will probably lose that argument.

Besides making it visible for developers who might be used to drupal 7, when you ask someone to reproduce their problem on the public demo if they don't see an error they might think it's their own setup and waste time.


demeritcowboy:
Thanks!

PR #547, opened 2020-08-22
cividist - When fetching remotes, use a fixed list of repos

totten:
For example: if you run `GIT_REMOTE=security cividist update`, and if
`vendor/pear/validate_finance_creditcard` is a git repo (*which it is*),
then it incorrectly tries to fetch the `security` remote on this repo.

Note that the previous call to `git scan` was fairly forgiving -- e.g. if the local source tree was fairly old, then it might not have (say) `drupal-8`. The behavior here is fairly close with/without the patch -- ie it updates the folder if present, and ignores it if missing.

### DIFF
--- a/bin/cividist
+++ b/bin/cividist
@@ -36,7 +36,15 @@ HASH_SUFFIX=-0003
 ###############################################################################
 ## Fetch the latest code
 function cividist_update() {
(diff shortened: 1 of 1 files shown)


seamuslee001:
This looks fine to me merging

PR #555, opened 2020-09-08
Increment minimum version - this prevents IDE warnings and brings us into 2016

eileenmcnaughton:
@totten

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
     "bin-dir": "bin"
   },
   "require": {
-    "php": ">=5.3.3",
+    "php": ">=5.4",
     "totten/php-symbol-diff": "dev-master#54f869ca68a3cd75f3386f8490870369733d2c23",
     "civicrm/upgrade-test": "dev-master#03e554787f378900f870a030cf2575d3cd49900f",
     "drupal/coder": "dev-8.x-2.x-civi#ef68b5617b94a1cef724b74c8b4318799388cc00",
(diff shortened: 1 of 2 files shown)


MikeyMJCO:
We should go further than this. Really we could/should peg this straight to `5.6`. [I already did for `cv`...](https://github.com/civicrm/cv/blob/master/composer.json)

composer.json
```diff
@@ -15,7 +15,7 @@
     "bin-dir": "bin"
   },
   "require": {
-    "php": ">=5.3.3",
+    "php": ">=5.4",
```
MikeyMJCO, line 18:
```suggestion
    "php": ">=5.6.0",
```

MikeyMJCO (changes requested):

composer.json
eileenmcnaughton: @MikeyMJCO I'd be happy with 7.1 - but at least this gets past the immediate issue...

@totten can you merge or merge the number you are happy to go to?

MikeyMJCO: So would I ;-)

5.6.0 aligns it with CV's minimum version (which is included in this composer.json)

totten: Yeah, I think 5.6 is good for now. (It's rare, but I did recently have to reproduce something where the configuration could only be initialized in PHP 5 + CIvi 4...) IMHO, we should bump this to php7 in a few months.

Also, the `composer.lock` needs  a small twiddling (`composer update --lock`).

Pushed up both those bits.

PR #558, opened 2020-09-09
Simplify core extensions testing

seamuslee001:
This aims to simplify the testing process for core extensions.  To allow for individual testing we still need to add lines into TestType case but this should be generic enough for PR / Matrix testing


seamuslee001:
@totten @eileenmcnaughton

PR #569, opened 2020-10-08
Enable search kit on demo sites

eileenmcnaughton:
This will enable us to share urls / info when investigating search options

### DIFF
--- a/app/config/drupal-demo/install.sh
+++ b/app/config/drupal-demo/install.sh
@@ -116,7 +116,7 @@ EOPERM
     add 'register to volunteer'
 EOPERM
 
-  cv en civirules civisualize cividiscount
+  cv en --ignore-missing civirules civisualize cividiscount org.civicrm.search
 
   ## Demo sites always disable email and often disable cron
   drush cvapi StatusPreference.create ignore_severity=critical name=checkOutboundMail
--- a/app/config/drupal8-demo/install.sh
+++ b/app/config/drupal8-demo/install.sh
@@ -97,7 +97,7 @@ pushd "${CMS_ROOT}/sites/${DRUPAL_SITE_DIR}" >> /dev/null
   drush8 -y rap anonymous 'register to volunteer'
   drush8 -y rap authenticated 'register to volunteer'
 
-  cv en civirules civisualize cividiscount
+  cv en --ignore-missing civirules civisualize cividiscount org.civicrm.search
 
   ## Demo sites always disable email and often disable cron
   drush8 cvapi StatusPreference.create ignore_severity=critical name=checkOutboundMail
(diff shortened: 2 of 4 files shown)


app/config/drupal-demo/install.sh
```diff
@@ -116,7 +116,7 @@ EOPERM
     add 'register to volunteer'
 EOPERM
 
-  cv en civirules civisualize cividiscount
+  cv en civirules civisualize cividiscount org.civicrm.search
```
seamuslee001, line 119: we may want to wrap this in an if statement cos not all versions of civicrm will have the search kit code available

eileenmcnaughton: what about an approach like this https://github.com/civicrm/civicrm-buildkit/pull/569/files#diff-f69ba7259c51f34bf3e1b9047bf97ca8R121

totten: * Yeah, doing some kind of batch selector is interesting.

    * There used to be support for selecting path (eg `cv api extension.install path=$PWD/ext/'*'`), but that's regressed.

    * Since we have 'tags' now, it might be better to [have a tag selector](https://github.com/civicrm/cv/pull/82), e.g.

        ```xml
        <tags>
          <tag>demo</tag>
        </tags>
        ```

        ```
        cv en @demo
        ```

* Wasn't the extension being renamed to `searchkit`?

* `cv en` has an option `--ignore-missing`. This may be actually be the simplest thing.

MikeyMJCO: +1 to "--ignore-missing"

Keep it simple :-)

eileenmcnaughton: ohh ignore missing = yes!

eileenmcnaughton:
@totten @MikeyMJCO @seamuslee001 @colemanw I've fixed it to use ignore-missing. 

I think it should be mergeable now but we should consider
1) taking the search renaming a bit further?
2) adding contact layout editor on demo sites

totten:
> > Wasn't the extension being renamed to searchkit?
>
> 1. taking the search renaming a bit further?

Aaaah, I forgot that it was already published as `org.civicrm.search` in v5.29, so I can see why maybe the rename didn't go further. That rename needs to be a whole other thing I guess.

From the buildkit POV, I imagine it'll be easy to accommodate the rename (when/if it finishes) by softly activating both names:

```
cv en --ignore-missing search searchkit
```

So we don't block here on that.

> 2. adding contact layout editor on demo sites

Yeah, that's a good idea too.

eileenmcnaughton:
@totten so ok to merge this? I will follow up with contact layout editor

colemanw:
FYI I don't see any reason to change the machine name of `org.civicrm.search` but if you disagree I can be overruled.

PR #570, opened 2020-10-10
[REF] Fix MySQLDump to not throw errors when dumping after MySQL 5.7.31

seamuslee001:
as per https://dba.stackexchange.com/questions/271981/access-denied-you-need-at-least-one-of-the-process-privileges-for-this-ope the recommended fix for this is to specify --no-tablespaces in the mysqldump command

@totten @MikeyMJCO


totten:
Looks like the simplest way to add the parameter to dumps. :+1:

Since the parameter is newly required on 5.7.31, the main question is whether it's safe on older versions. I did a quick trial run on 5.6.47 (ie `civibuild reinstall dmaster` ==> muck with manually ==> `civibuild restore dmaster`), and the new parameter didn't seem to cause any problems.

Merging.

PR #571, opened 2020-10-11
Enable org.civicrm.contactlayout on demo sites

eileenmcnaughton:
@totten @colemanw @seamuslee001 @MikeyMJCO  same as https://github.com/civicrm/civicrm-buildkit/pull/569 for contact layout editor.

we already enable civirules & civivolunteer so I would consider mosaico too - but not in scope on this


seamuslee001:
+:+1:

PR #578, opened 2020-11-11
Add PHPStan CiviCRM

mglaman:
Added as a download due to PHPStan requiring 7.1 or greater. 

The phar can be executed against any code, still, and configured to point at the phpstan-civicrm extension.

The main task is creating a ./bin/civi-phpstan runner which locates the $civicrm_root and generates a phpstan.neon for the working directory. Thus far the extension for PHPStan has been tested against Drupal 8/9 Composer builds.


mglaman:
Actually, as I think of this, I do not believe it belongs here. This is something that would be executed directly against `civicrm-core`. But I'm not used to developer workflows here. For example, for static analysis of CiviCRM extensions can you do that against the `civicrm-core` or does it require an install with a CMS.

Basically, in order to run PHPStan CiviCRM, it needs to know where the $civicrm_root is located. And such buildkit doesn't allow installing PHPStan directly as a dependency due to minimum PHP requirements, this needs to run 

```
php ./extern/phpstan/phpstan.phar
```

A generated `phpstan.neon` will need to then load the extension in `./extern/phpstan/ext-civicrm/extension.neon` and the defined civicrm_root parameter.

JoeMurray:
@seamuslee001 can you look at this next week, please?

PR #581, opened 2020-11-16
Enable searchkit on drupal-clean

eileenmcnaughton:
I have focussed on drupal clean as it is used on PR sites - however eventually this will be in
the main install script and enabled by default on all new sites. This is the next step on that
core-extension path

@totten @colemanw @seamuslee001


colemanw:
This is ok with me. @totten?

eileenmcnaughton:
@colemanw merge?

PR #584, opened 2020-11-25
Fix patching for MySQL 8 on Drupal 7 by patching for any CMS version …

seamuslee001:
…that starts with 7 and updating the grep based on the merged in patch

I tested this locally and it correctly doesn't apply the patch on the current d7 dev but does on 7.74 which is the most recent release of Drupal 7


seamuslee001:
Merging as per comment on infra that Tim is ok with me to self merge here

PR #585, opened 2020-12-11
The repo for drupal 8 is way old

demeritcowboy:
Also civicrm-setup is in core now.

I've left all the new docs repos out of this PR, and there's probably other possible updates. Mostly looking at drupal 8 here.


totten:
Looks good to me.

demeritcowboy:
Thanks!

PR #586, opened 2020-12-15
civi-ssh-host-keys - Add helper script to setup known_hosts

totten:
This script registers the public keys for a couple SSH hosts (`lab.civicrm.org`, `github.com`) if they are not already authorized. It specifically uses concrete host keys because:

* The keys don't change often, and..
* The typical 'trust-on-first-use' (TOFU)  implementations are not reliable across different CI environments.


seamuslee001:
Makes sense Merging

PR #587, opened 2020-12-21
Include STRICT_TRANS_TABLES in the SQL_Modes by default for MySQL 5.6…

seamuslee001:
…, MySQL 5.7,8 already has it in default

@totten @demeritcowboy

### DIFF
--- a/.loco/config/mysql-common/my.cnf.php
+++ b/.loco/config/mysql-common/my.cnf.php
@@ -38,12 +38,13 @@ innodb_file_per_table = 1
 #innodb_lock_wait_timeout = 50
 
 <?php if (matchVer('/Ver 5.6/')) { printf("innodb_large_prefix = 1\n"); } ?>
+<?php if (matchVer('/Ver 5.6/')) { printf("sql-mode=\"STRICT_TRANS_TABLES,NO_ENGINE_SUBSTITUTION\"\n"); } ?>
 <?php if (!matchVer('/Ver 8.0/')) { printf("innodb_file_format = Barracuda\n"); } ?>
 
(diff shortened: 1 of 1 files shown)


.loco/config/mysql-common/my.cnf.php
```diff
@@ -38,12 +38,13 @@ function matchVer($pat) {return (bool) preg_match($pat, ver());}
 #innodb_lock_wait_timeout = 50
 
 <?php if (matchVer('/Ver 5.6/')) { printf("innodb_large_prefix = 1\n"); } ?>
+<?php if (matchVer('/Ver 5.6/')) { printf("sql_mode=\"STRICT_TRANS_TABLES,NO_ENGINE_SUBSTITUTION\"\n"); } ?>
```
demeritcowboy, line 41: My experience is this isn't additive, i.e. this would overwrite the default not add to it. But I'm not familiar with the environment this runs under.

seamuslee001: by default MySQL 5.6 is only NO_ENGINE_SUBSITUTION this adds to it by including STRICT_TRANS_TABLES

demeritcowboy: Ah ok got it. 👍

eileenmcnaughton:
@demeritcowboy is this mergeable now? Seems OK to me

demeritcowboy:
I don't really use buildkit (cuz windows), but my initial lack of knowledge about 5.6 is rectified so no objections.

eileenmcnaughton:
OK - I'm going to merge - any issues will turn up in tests but I don't expect to see them

PR #588, opened 2020-12-28
NO_AUTO_CREATE_USER removed upstream

ufundo:
Removing the NO_AUTO_CREATE_USER option for Mysql 8 has made it's way into the upstream drupal source https://api.drupal.org/api/drupal/includes!database!mysql!database.inc/7.x 🙌

As a result the first hunk of this patch is no longer needed (and throws an error when building currently)

### DIFF
--- a/app/drupal-patches/mysql8-drupal.patch
+++ b/app/drupal-patches/mysql8-drupal.patch
@@ -2,15 +2,6 @@ diff --git a/includes/database/mysql/database.inc b/includes/database/mysql/data
 index 356e039f..c803207d 100644
 --- a/includes/database/mysql/database.inc
 +++ b/includes/database/mysql/database.inc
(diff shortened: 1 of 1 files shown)


seamuslee001:
@ufundo have you pulled the latest version of buildkit down as this should already be addressed in other changes

ufundo:
@seamuslee001 eek, no sorry and that does indeed resolve the issue. is it that this patch file is no longer used during the install? can it be removed maybe?

seamuslee001:
@ufundo yes but it is still needed if you decided to pick an earlier version of the CMS which is why its now been made conditional in the patch https://github.com/civicrm/civicrm-buildkit/commit/469a36fccadc7ab586bcf208079b8863bf17073f

ufundo:
ah that makes sense also, apologies I'm new here!

PR #590, opened 2021-01-05
Upgrade MySQL 8.0 version to one that might be a bit more performant

seamuslee001:


nix/pkgs/mysql80/8.0.x.nix
```diff
@@ -0,0 +1,77 @@
+{ lib, stdenv, fetchurl, bison, cmake, pkgconfig
+, boost, icu, libedit, libevent, lz4, ncurses, openssl, protobuf, re2, readline
+, zlib, zstd, numactl, perl, cctools, CoreServices, developer_cmds, libtirpc, rpcsvc-proto
```
totten, line 3: When I tried this locally, it failed with:

```
totten@zax:~/bknix$ nix-shell -A edge
error: anonymous function at /home/totten/bknix/nix/pkgs/mysql80/8.0.x.nix:1:1 called without required argument 'rpcsvc-proto', at /home/totten/bknix/nix/pkgs/mysql80/default.nix:9:8
(use '--show-trace' to show detailed location information)
```

I don't think that `rpcsvc-proto` is available in 20.09:

```
totten@zax:~/src/nixpkgs$ git checkout 20.09
totten@zax:~/src/nixpkgs$ grep -r rpcsvc-proto |wc
      0       0       0

totten@zax:~/src/nixpkgs$ git checkout cd63096d6d887d689543a0b97743d28995bc9bc3 
totten@zax:~/src/nixpkgs$ grep -r rpcsvc-proto |wc
      0       0       0

totten@zax:~/src/nixpkgs$ git checkout master
totten@zax:~/src/nixpkgs$ grp -r rpcsvc-proto |wc
     34     212    3205
```

seamuslee001: should be sorted now @totten not sure where that came from

totten:
Note: We discussed further at https://chat.civicrm.org/civicrm/pl/wno9gcjy43nyiytgfqpn7fo4nc

@seamuslee OK, so I extended nixpkgs#93371 and submitted https://github.com/NixOS/nixpkgs/pull/109006

It feels like the pattern of copying in a package (for backporting/forking purposes), and maybe we should do something that keeps the code a bit closer to upstream/nixpkgs, e.g.

* Add/use a pin for the PR branch -- like https://github.com/civicrm/civicrm-buildkit/pull/593
* Add a fork `civicrm/nixpkgs` based on `NixOS/nixpkgs`. Maintain a branch in there with cherry-picked PRs and periodically rebase -- like https://gist.github.com/totten/39e932e5d10bc9e73e82790b2475eff2

seamuslee001:
I like your alternate PR looks right to me closing in favour of your work

PR #591, opened 2021-01-11
joomlatools-console - Switch from custom PHAR to extern

totten:
__Overview__: `joomlatools-console` provides a subcommand for installing Joomla.

__Before__: `civi-download-tools` downloads a PHAR build of `joomlatools-console`. However, this PHAR is based on a fork that is not maintained.

__After__: `civi-download-tools` creates a folder `extern/joomlatools-console` based on a git tag of `joomlatools-console`. This aligns better with upstream's distribution, and it avoids impedance mismatch between their composer.json and buildkit's composer.json.


seamuslee001:
Looks fine to me merging

PR #595, opened 2021-02-01
Remove database.inc patch about NO_AUTO_CREATE_USER

artfulrobot:
It looks like some php logic has been implemented such that the first part of this patch (I'm referring to a patch file saved in the project, not a commit) might not be necessary any longer.

Put another way: with buildkit as is, I couldn't get it to build because it failed trying to apply this patch. With this PR, I could get it to build.

### DIFF
--- a/app/drupal-patches/mysql8-drupal.patch
+++ b/app/drupal-patches/mysql8-drupal.patch
@@ -2,15 +2,6 @@ diff --git a/includes/database/mysql/database.inc b/includes/database/mysql/data
 index 356e039f..c803207d 100644
 --- a/includes/database/mysql/database.inc
 +++ b/includes/database/mysql/database.inc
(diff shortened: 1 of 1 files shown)


seamuslee001:
@artfulrobot I think we should keep the patch around in case you specified a cms version that was lower than the current release did your local patch have this? https://github.com/civicrm/civicrm-buildkit/commit/469a36fccadc7ab586bcf208079b8863bf17073f

seamuslee001:
@artfulrobot did you make sure your image was up to date? just wondering if it was using a cached version of the buildkit folder

michaelmcandrew:
@artfulrobot - feel free to give me a shout before you embark on the next 7 hours. It might be that the [documentation](https://github.com/michaelmcandrew/civicrm-buildkit-docker) is missing something. We might be able to improve it for the next person (the next person including you :))

PR #610, opened 2021-03-20
Use PHP7.4 for Max profile

seamuslee001:
@totten we now have a fully passing build on edge https://test.civicrm.org/job/CiviCRM-Core-Edge/CIVIVER=master,label=test-3/ this will be the next step IMO. ping @eileenmcnaughton 

I am thinking that we may want to hold back for another couple of release cycles at least until 5.37 becomes stable release.

I also started looking at PHP8 and it seems like there might have been some movements on 2 of the 3 pecl packages but not timecop

### DIFF
--- a/nix/profiles/max/default.nix
+++ b/nix/profiles/max/default.nix
@@ -11,7 +11,7 @@ let
 
 in (import ../base/default.nix) ++ (import ../mgmt/default.nix) ++ [
 
-    bkpkgs.php73
+    bkpkgs.php74
     pkgs_1809.nodejs-8_x
     pkgs.apacheHttpd
     pkgs_1809.mailcatcher


eileenmcnaughton:
@seamuslee001 so we can make 7.4 supported AND recommended from 5.37

eileenmcnaughton:
hmm wanna wait another month?

totten:
To my read of https://test.civicrm.org/job/CiviCRM-Core-Edge/, it looks like `master` and `5.37` only have flaky-test failures, but 5.36 still has massive failures.

I'd say let's merge/deploy as soon 5.37 supercedes 5.36 (circa May 5).

eileenmcnaughton:
yay

PR #612, opened 2021-04-26
Ensure that we install search kit no matter the civi version if it is…

seamuslee001:
… appropriate

ping @eileenmcnaughton @colemanw


app/config/drupal-clean/install.sh
```diff
@@ -67,6 +67,6 @@ pushd "${CMS_ROOT}/sites/${DRUPAL_SITE_DIR}" >> /dev/null
     drush -y role-add-perm civicrm_webtest_user "$perm"
   done
 
-  cv en --ignore-missing org.civicrm.search
+  #cv en --ignore-missing org.civicrm.search_kit org.civicrm.search
```
colemanw, line 70: @seamuslee001 so you're disabling it on drupal-clean?

PR #620, opened 2021-05-19
(REF) app/config/*-demo - Get list of demo exts from common env-var

totten:
This consolidates the list of extensions used on various `*-demo` builds.

The demo builds already use the same list of extensions, although the code
from each `install.sh` has some arbitrary differences.  (In particular, some
scripts use multiple calls to `drush cvapi` and `cv en`; other scripts use
a single `cv en`. This is just evolutionary happenstance.)

Two upshots:

1. When changing the `*-demo` cfg, you only have to change that part once.

2. If you want to override this list locally, you can edit `civibuild.conf` and set `CIVI_DEMO_EXTS`.


seamuslee001:
Looks good to me

PR #626, opened 2021-05-22
PHPUnit8 Remove warnings attribute from phpunit testsuite to work wi…

seamuslee001:
…th current version of xunit

ping @totten @eileenmcnaughton

### DIFF
--- a/bin/phpunit-xml-cleanup
+++ b/bin/phpunit-xml-cleanup
@@ -16,5 +16,6 @@ foreach ($files as $file) {
   $content = preg_replace('/(<testsuite.*) fullPackage="[^"]+" /', '$1 ', $content);
   $content = preg_replace('/(<testsuite.*) subpackage="[^"]+" /', '$1 ', $content);
   $content = preg_replace('/(<testsuite.*) namespace="[^"]+" /', '$1 ', $content);
+  $content = preg_replace('/(<testsuite.*) warnings="[^"]+" /', '$1 ', $content);
   file_put_contents($file, $content);
 }
--- a/src/command/env-info.run.sh
+++ b/src/command/env-info.run.sh
@@ -52,6 +52,8 @@ env_report_cmd php-fpm --version
 env_report_cmd phpunit5 --version
 env_report_cmd phpunit6 --version
 env_report_cmd phpunit7 --version
+env_report_cmd phpunit8 --version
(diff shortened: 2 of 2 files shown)


eileenmcnaughton:
can you explain this one a bit more?

seamuslee001:
XUnit which is the tool that process the xml results fails if it receives a xml file that doesn't match its standards. PHPUnit8 added warnings metric into the xml output from test runs. the version of XUnit that is currently installed on Jenkins doesn't support the warnings tag see conversation here https://chat.civicrm.org/civicrm/pl/5bzdr7xrqtdjmp54f4fpkqqm5a

seamuslee001:
This removes the warnings metric which makes Xunit be able to process the generated result file

eileenmcnaughton:
@seamuslee001 so the options are upgrade xunit or do this - did you look into the former ?

seamuslee001:
@eileenmcnaughton I did but i'm nervous about doing it as we would need to go from 1.99 to 3.0.2 and it warns "Warning: The new version of this plugin is marked as incompatible with the installed version. This is usually the case because its behavior changed, or because it uses a different settings format than the installed version. Jobs using this plugin may need to be reconfigured, and/or you may not be able to cleanly revert to the prior version without manually restoring old settings. Consult the plugin release notes for details." I don't know what testing regime we have for doing jenkins plugin updates

eileenmcnaughton:
So what are these warnings we are suppressing - it feels like if we suppress them now there is an odds-on chance we will never unsuppress them

eileenmcnaughton:
Also - add the link here https://plugins.jenkins.io/xunit/#documentation -  it looks like version 2 is the one with the breaking changes?

seamuslee001:
@eileenmcnaughton we are suppressing the warnings counter within the junit-xml i.e. we are moving from the top line of each test suites, we are removing the warnings key here

<name="xxx" failures="xxx" passes="xxx" warnings="0">

seamuslee001:
Yes I think its and I didn't just want to go ahead and upgrade the plugin in case it broke our jenkins setup (I have no idea if it will or not)

eileenmcnaughton:
@seamuslee001 OK that warnings counter seems fairly trivial then - I think that by merging this I'm accepting we will never remember to re-enable it but I guess it's ok....

PR #631, opened 2021-06-02
Update the version of loco shipped for php8 support

seamuslee001:
ping @totten


seamuslee001:
I tested this on test2 by manually applying the patch and reset the system and it correctly built

Issue #633, opened 2021-06-28
Node.js is outdated in download tools


wmortada:
`civi-download-tools` installs Node.js version 10, which reached EOL in April 2021.

E.g. see [line 502](https://github.com/civicrm/civicrm-buildkit/blob/fee9a80114c10f541a43a57179664fb8dc19fe8b/bin/civi-download-tools#L502)

```
      wget -O- https://deb.nodesource.com/setup_10.x | sudo -E bash -
```

This generates an error message on installation.

I think this should be fixed by updating this to version 14, but wasn't sure if this would cause other issues?

See also [related ticket](https://github.com/michaelmcandrew/civicrm-buildkit-docker/issues/68) for CiviCRM Docker Buildkit.

homotechsual:
That's for Debian Stretch - really tooling should run against Buster which should install NodeJS 12 or 14.

wmortada:
Aha, then perhaps Michael's docker set up should be using Buster too. I'll raise an issue there.

totten:
I'm +1 on moving to NodeJS 12 or 14. (Preferably v14, if it works...)

Note that, historically, previous updates to NodeJS were held back so that it would be possible to build Shoreditch on the default env. Of course, it only matters if you're actually patching that extension, and (in any event) Shoreditch has also [moved to NodeJS 14](https://github.com/civicrm/org.civicrm.shoreditch/commit/855dcc0af86adfd9952c6e509dd6829dbf6665bc).

#634 would bump up the version used for [edge-testing](https://test.civicrm.org/job/CiviCRM-Core-Edge/) in CI.

homotechsual:
v14 works - but I want to do some more in-depth testing on Wednesday.

wmortada:
It looks like [the default version in Debian Buster](https://packages.debian.org/buster/nodejs) is Node.js 10 so we'd need to use the nodesource script as above if we wanted to include a more recent version.

michaelmcandrew:
In case anyone is interested, we have stopped downloading our own version of node in civicrm-buildkit-docker and are now relying on civi-download-tools to do this for us. https://lab.3sd.io/tools/civicrm-buildkit-docker/-/commit/3b7531d7bf221021ac44756ab5d3cd2919bf22c9.

PR #634, opened 2021-06-28
nix/profiles/edge - Bump NodeJS from v8 to v14

totten:
If this works on `edge`, then we should bump up the {`max`,`dfl`} profiles as well.

See also: #633

CC @seamuslee001


seamuslee001:
Tested this on edge and looks fine to me

Issue #635, opened 2021-07-05
Error running civi-download-tools


wmortada:
I'm getting the following error when I run `civi-download-tools`:

```
33 error code EBADPLATFORM
34 error notsup Unsupported platform for fsevents@2.3.2: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
35 error notsup Valid OS:    darwin
35 error notsup Valid Arch:  any
35 error notsup Actual OS:   linux
35 error notsup Actual Arch: x64
```

After doing a bit of digging, I think it is because `fsevents` is Mac only and I'm not on a Mac.

I fixed the issue by deleting `package-lock.json`. This file was [recently added](https://github.com/civicrm/civicrm-buildkit/commit/df2452e803cc882332b095eaf726a0b1f146a116#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519) to git by @seamuslee001. I'm wondering if this was a mistake?

wmortada:
Fixed by #636

PR #636, opened 2021-07-05
#635 Fix Package Lock for linux

seamuslee001:
@wmortada can you try this out please?


wmortada:
Yup that works for me. Thanks @seamuslee001.

PR #642, opened 2021-07-27
Dump stored procedures too when making snapshot

demeritcowboy:
Overview
----------
This is a bit speculative, but my theory is this is the cause of https://lab.civicrm.org/dev/core/-/issues/2692 and I can't think of a downside, and may help devs in a more general case.

BEFORE
---------
mysqldump will dump triggers by default but not stored procedures. Civi has civicrm_strip_non_numeric, so when reloading the snapshot, you lose the procedure.

AFTER
-------
Procedures are dumped and so also get reloaded if the db is restored.

Comments
------------
Another option would be for people's scripts to rebuild the triggers which recreates the stored procedure. But it seems more consistent to always dump what was already existing.


eileenmcnaughton:
ok - that should be non-controversial - let's try it

demeritcowboy:
Thanks. I don't know if the test nodes pick it up automatically - buildkit might be preinstalled there.

PR #654, opened 2021-11-08
Update Max Profile to use php8.0 and dfl (PR Testing) to use php7.4 w…

seamuslee001:
…hilst keeping php7.2 on min

ping @eileenmcnaughton @demeritcowboy @totten

### DIFF
--- a/nix/profiles/dfl/default.nix
+++ b/nix/profiles/dfl/default.nix
@@ -12,7 +12,7 @@ let
 
 in (import ../base/default.nix) ++ (import ../mgmt/default.nix) ++ [
 
-    bkpkgs.php72
+    bkpkgs.php73
     pkgs_2105.nodejs-14_x
     pkgs_2105.apacheHttpd
     pkgs_1809.mailcatcher
--- a/nix/profiles/dfl/default.nix
+++ b/nix/profiles/dfl/default.nix
@@ -12,7 +12,7 @@ let
 
 in (import ../base/default.nix) ++ (import ../mgmt/default.nix) ++ [
 
-    bkpkgs.php73
+    bkpkgs.php74
     pkgs_2105.nodejs-14_x
     pkgs_2105.apacheHttpd
     pkgs_1809.mailcatcher
--- a/nix/profiles/max/default.nix
+++ b/nix/profiles/max/default.nix
@@ -12,7 +12,7 @@ let
 
 in (import ../base/default.nix) ++ (import ../mgmt/default.nix) ++ [
 
-    bkpkgs.php74
+    bkpkgs.php80
     pkgs_2105.nodejs-14_x
     pkgs_2105.apacheHttpd
     pkgs_1809.mailcatcher


totten:
Looks like `edge` (`php80`) is still raising errors on `civicrm-core@5.44` and `civicrm-core@master`?

https://test.civicrm.org/job/CiviCRM-Core-Edge/CIVIVER=5.44,label=test-3/lastCompletedBuild/testReport/
https://test.civicrm.org/job/CiviCRM-Core-Edge/CIVIVER=master,label=test-3/lastCompletedBuild/testReport/

We should merge as soon as `php80` is passing.

IMHO, it would be safe to go ahead with bumping up `dfl` to `php74`. But if you wanted to do that first, then the changes should be split to separate PRs.

seamuslee001:
@totten those are only recently created errors due to recent PR changes

eileenmcnaughton:
I think 'still' is the wrong word....

totten:
Yeah, "still" is the wrong word. 

I guess my point is this - if we flip the main matrix to php80 while it appears to report 30+ failures, then it'll teach us to ignore the main matrix. 😟 

It'll be curious to see if php80 turned red due to Smarty issues? That category of issue seems to be super-whack-a-moley. If we have gotten pulled into a loop of Smarty/php80 interactions, then I might come round to @eileenmcnaughton's view on doing PR tests with php80. Or maybe we take another stab at doing multiple test-jobs on PRs. ([This commenter](https://github.com/janinko/ghprb/issues/142#issuecomment-370061915) and the :+1:s seems to indicate that it may work now, although it's curious because the issue is still listed as open...)

demeritcowboy:
I think the last half or so are that thing where when a test involving smarty fails at a certain point smarty's secure mode doesn't get reset, so then everything fails. I forget where in smarty that happens but it's come up a few times before.

It seems like we mainly need to deal with this one: https://github.com/civicrm/civicrm-core/pull/21979

> If we have gotten pulled into a loop of Smarty/php80 interactions

A little bit probably. It wouldn't surprise me if there will be more coming.

totten:
> I think the last half or so are that thing where when a test involving smarty fails at a certain point smarty's secure mode doesn't get reset, so then everything fails.

Can we reset the secure mode in `CiviTestListener` or `CiviUnitTestCase`? Or I wonder if it would kill performance to reset `CRM_Core_Smarty::$_singleton` completely? If we're liable to get more failures in that vein in the future, then deobfuscating the error seems useful...

demeritcowboy:
Yep I just put up a PR that might deal with that if it doesn't break stuff https://github.com/civicrm/civicrm-core/pull/22012

eileenmcnaughton:
The last fails are all introduced by the changes to the smarty templates - which introduced enotices in the tests (while removing them from the UI) -  which only phpunit 8 set up  cares about, The secure mode issue is a symptom- but solving that doesn't solve the fact that if we continue with the work to enable escape on output we are gonna keep breaking php 8 tests.

 I had been wondering if maybe that was a case for discontinuing the work on escape-on-output for a while (e.g 6-12 months) while our php versions move up a bit

seamuslee001:
We discussed this on the CT call and decided that we would switch PR testing to be on php8 and that we would mark tests as icomplete/ornery if we can't get them fixed quickly

demeritcowboy:
For the Session::setStatus/info.tpl failing test, that was one I added for an unrelated reason, and I can put up a PR for that one since I think it can just be dealt with in the test. What it was really just testing is that if you leave out the message body and just have a title it still pops up an alert, and then I went all out on it since it was just some easy copy/paste.

For the smarty secure_mode/static, it wasn't obvious why my local runs seem so different from the test nodes, so if I don't have time to look into that I might close that - or try just having CiviUnitTestCase reset secure_mode like @totten mentioned earlier.

eileenmcnaughton:
Ok - I'm gonna merge this per CT call - we just have to fix the tests quicky

PR #656, opened 2021-11-20
Fix installation of backdrop using latest drush8 dev version on php8

seamuslee001:
@herbdool @totten this seemed to get around the bootstrapping problem on php8 and also updates to latest tagged release of drush-backdrop

### DIFF
--- a/bin/civi-download-tools
+++ b/bin/civi-download-tools
@@ -743,10 +743,6 @@ pushd $PRJDIR >> /dev/null
   make_link "$PRJDIR/bin" "../extern/phpunit8/phpunit8.phar" "phpunit8"
   make_link "$PRJDIR/bin" "../extern/phpunit9/phpunit9.phar" "phpunit9"
 
-  [ ! -d "$HOME/.drush" ] && mkdir "$HOME/.drush"
-  [ ! -d "$HOME/.drush/command" ] && mkdir "$HOME/.drush/command"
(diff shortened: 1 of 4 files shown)


bin/civi-download-tools
```diff
@@ -743,6 +743,10 @@ pushd $PRJDIR >> /dev/null
   make_link "$PRJDIR/bin" "../extern/phpunit8/phpunit8.phar" "phpunit8"
   make_link "$PRJDIR/bin" "../extern/phpunit9/phpunit9.phar" "phpunit9"
 
+  [ ! -d "$HOME/.drush" ] && mkdir "$HOME/.drush"
+  [ ! -d "$HOME/.drush/command" ] && mkdir "$HOME/.drush/command"
+  make_link "$HOME/.drush/command" "$PRJDIR/extern/drush-lib/backdrop" "backdrop"
```
totten, line 748: Currently, `bin/drush8` is [a wrapper that applies `--include="$PRJDIR/extern/drush-lib"`](https://github.com/civicrm/civicrm-buildkit/blob/master/src/drush/drush8.tmpl). This allows the `buildkit` directory to be self-contained (ie you can clone/run/delete without any effect on system state; you can have multiple copies without affecting each other).

Did something change in `drush` or `drush-backdrop` to make `HOME` mandatory?

seamuslee001: @totten no i just seemed to manage to clobber the wrapper completely in my testing of drush test phars it seems

PR #661, opened 2021-12-21
Use CV to do install on drupal-clean installs

seamuslee001:
@totten this seems to work for me

### DIFF
--- a/app/config/drupal-clean/download.sh
+++ b/app/config/drupal-clean/download.sh
@@ -20,6 +20,9 @@ pushd "$WEB_ROOT/web"
 
     extract-url --cache-ttl 172800 civicrm=http://download.civicrm.org/civicrm-l10n-core/archives/civicrm-l10n-daily.tar.gz
     ## or https://raw.github.com/civicrm/l10n/master/po/fr_CA/civicrm.mo => civicrm/l10n/fr_CA/LC_MESSAGES/
+    pushd civicrm
+      composer install
+    popd
   popd
 
 popd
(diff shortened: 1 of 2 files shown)


app/config/drupal-clean/download.sh
```diff
@@ -20,6 +20,9 @@ pushd "$WEB_ROOT/web"
 
     extract-url --cache-ttl 172800 civicrm=http://download.civicrm.org/civicrm-l10n-core/archives/civicrm-l10n-daily.tar.gz
     ## or https://raw.github.com/civicrm/l10n/master/po/fr_CA/civicrm.mo => civicrm/l10n/fr_CA/LC_MESSAGES/
+    pushd civicrm
+      composer install
```
totten, line 24: Aaah, right - `civicrm_install` calls `setup.sh -Dgsef` (or `setup.sh -Dgsdf`), and `civicrm_install_cv` doesn't call `setup.sh` at all. When switching, we miss the equivalent of `setup.sh -Dg`.

Just to verbalize... in practical terms, this means that `drupal-clean` will skip these things:

* `npm install`
* `GenCode`

And... that seems like a good thing! Like... it wouldn't have worked in (say) Civi 5.0, but in the post-D8/post-composerization, this seems much more realistic.

> this seems to work for me

Does that mean on a local build? Or on a `test-N` box? It might make sense to deploy this to one of the test nodes and let it run some jobs.

totten: You know, for some # months, every time I run `civibuild reinstall`, it updates `package-lock.json`. To avoid superfluous changes, I have to frequently run `git checkout -- package-lock.json`. And it sounds like that would go away when we make this change. :smile:

seamuslee001: @totten it meant works for me on a local build. but also we should be keeping package-lock.json up to date like the composer.lock equivalent

PR #662, opened 2021-12-21
dev/core#2998 Towards CiviCRM Standalone support

mlutfy:
It works up to the point where it call `cv`, which cannot detect the CMS.

I then proceeded to complete the install by:

* Fetching the DB credentials from `~/.amp/instances.yml`
* `chgrp -R www-data builds/mysite/data`
* `chmod -R g+w builds/mysite/data`
* Complete the setup using the web UI

There are some "todos":

- we cannot `composer create-project` yet (package should be registered with packagist?)
- there is no user management yet

cc @demeritcowboy @totten @homotechsual

### DIFF
--- a/app/config/standalone-clean/download.sh
+++ b/app/config/standalone-clean/download.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+## download.sh -- Download Standalone project and CiviCRM core
+
+###############################################################################
+
+# @todo Requires that we register with packagist?
+# mkdir "$WEB_ROOT"
+# composer create-project mlutfy/civicrm-standalone "$WEB_ROOT" --no-interaction
+git clone https://github.com/mlutfy/civicrm-standalone $WEB_ROOT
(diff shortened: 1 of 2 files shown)


mlutfy:
I'm not sure what that means :)

mattwire:
> I'm not sure what that means :)

Gone hard like bread and no good for eating anymore :-)

(This repo has an auto-triage function setup to flag stuff that might be abandoned)

totten:
Let's merge this. I think it's better in than out. There may be some TODOs, but that sounds like a chicken-egg situation. (If one wanted to make a more cogent build by working on patches for user-mgmt or for `cv`, then how would you initialize the build?) 

IMHO, it's OK+expected that the standalone build would be in a sort of WIP/prealpha stage for a while, and there's no downside to merging quickly on changes for `app/config/standalone-clean`.

mlutfy:
@totten Thanks!

PR #664, opened 2021-12-22
fix add_wpload_setting - to prevent fatal errors in E2E testing

kcristiano:
@christianwach pointed out that he saw  fatal errors on PR builds due to the way we call  `add_wpload_setting`   I've updated to the sysntax that is currently working for me locally.


seamuslee001:
if this works for you @kcristiano I'm happy to go with it

PR #665, opened 2022-01-05
dev/core#2773 Add a new user to the drupal demo install 'advisor'

eileenmcnaughton:
This creates a new user 'advisor' password 'advisor' with the email jenny@example.com.

In conjunction with https://github.com/civicrm/civicrm-core/pull/22377 this new login gets ACL limited
access to contacts rather than 'view all' - this makes testing easier

Note the key thing I was trying to set up is an ACL limited user. The user also only has access to view contributions - but this could morph over time if we find other configs more useful

image

### DIFF
--- a/app/config/drupal-demo/install-welcome.php
+++ b/app/config/drupal-demo/install-welcome.php
@@ -25,7 +25,7 @@ Just login with...
 ... and click the <a href="/civicrm"><strong>CiviCRM</strong></a> link (upper left-hand corner of your screen)
 
 <p><strong>Any data you enter on this demo site is "publicly available" due to the open login. Please do not enter real email addresses or other personal information.</strong> The demo database is reset periodically.</p>
-
+<p>There is also a user with the username advisor and the password advisor who has more limited access granted by Access Control</p>
 <h3>New to CiviCRM?</h3>
 <ul>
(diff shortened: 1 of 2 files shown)


kcristiano:
flagging for inclusion in `wp-demo` as well.   I will try and get a matching PR.

eileenmcnaughton:
@seamuslee001 @colemanw @totten should we get this merged to improve our ability to test (working well locally but I expect it won't be the last word on the permissions for the new user)

colemanw:
Ok with me.

totten:
I'm OK with adding the user for `*-demo`, and I really like have some sample ACLs for `r-run`ning.

`civibuild` has some strong opinions about users 1+2 (eg it respects/sets/gets `ADMIN_USER` and `DEMO_USER` - and makes these available for phpunit scripts).  There are potential issues around adding hardcoded usernames into builds, but there are diminishing returns in abstracting extra users (`FOO_USER`, `BAR_USER`, ...).  IMHO, `*-demo` is an appropriate place to say, "We can hardcode some usernames into this build-type."

Maybe I'm OCD, but the intro-text feels a little asymmetric. It should list different accounts (`demo/demo`, `advisor/advisor`) with similar formatting.

eileenmcnaughton:
@totten you can always fix the intro text to how you like it :-)

PR #666, opened 2022-01-09
Give demo user permission to close & edit own batches

eileenmcnaughton:
As an aside they already have this in d8 & still don't in wordpress. At some point
we should use a config file that they all read for users, roles & permissions
but it's a bit of a lift.

### DIFF
--- a/app/config/drupal-demo/install.sh
+++ b/app/config/drupal-demo/install.sh
@@ -122,6 +122,8 @@ EOPERM
     add 'access CiviGrant'
     add 'edit grants'
     add 'delete in CiviGrant'
+    add 'close own manual batches'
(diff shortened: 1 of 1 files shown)


totten:
I looked at this a couple days ago and wanted to merge after two things:

* `r-run`
* Check the permissions exist recent versions  (eg 5.39 )

I did both of these today, and it looks like these perms go back quite a ways. So if you still think this is a good idea, let's reopen+merge.

eileenmcnaughton:
Hmm - looks like I already deleted the branch after it got closed - but seems easy to redo

eileenmcnaughton:
OK https://github.com/civicrm/civicrm-buildkit/pull/675

PR #667, opened 2022-01-11
Ensure that CiviGrant extension is enabled on demo sites and appropri…

seamuslee001:
…ate permissions are granted as applicable

@eileenmcnaughton @colemanw

### DIFF
--- a/app/config/drupal8-demo/install.sh
+++ b/app/config/drupal8-demo/install.sh
@@ -38,7 +38,7 @@ pushd "${CMS_ROOT}/sites/${DRUPAL_SITE_DIR}" >> /dev/null
   drush8 cc drush -y
 
   ## Setup CiviCRM
-  echo '{"enable_components":["CiviEvent","CiviContribute","CiviMember","CiviMail","CiviReport","CiviPledge","CiviCase","CiviCampaign"]}' \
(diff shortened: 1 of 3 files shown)


eileenmcnaughton:
Agree with this - as long as it works (ie the extension is enabled before the permissions are added)

D7 = missing ?

My main 'wait for a bit' is just that I think we should sort out the existing permission inconsistencies first (since I hit some on the demo site in testing)

seamuslee001:
@eileenmcnaughton D7 is already granting the permissions necessary https://github.com/civicrm/civicrm-buildkit/blob/master/app/config/drupal-demo/install.sh#L122 however this is only done on the -demo build, PR tests spin up a drupal-clean build not drupal-demo same in wp-demo already https://github.com/civicrm/civicrm-buildkit/blob/master/app/config/wp-demo/install.sh#L120

eileenmcnaughton:
@seamuslee001 hmm - that's a pain - it makes it hard to test the pr site

seamuslee001:
@eileenmcnaughton I note that dmaster has been failing to build because of this https://test.civicrm.org/view/Sites/job/demo.civicrm.org/BLDNAME=dmaster,label=www-demo/2880/console any objections to me merging?

Issue #663, opened 2021-12-21
ensure composer runs php7.3 in debian bullseye


jmcclelland:
In `civi-download-tools`, when the script detects Debian bullseye, we install php7.3 packages (using the sury repository).

However, the default php version in bullseye is 7.4. So, when composer runs via the command line, it runs php 7.4, and it complains about a missing mbstring php package.

The simple fix is to add:

`sudo update-alternatives --set php /usr/bin/php7.3`

But... I think a better fix is to remove sury, remove the php version number from the packages (to bring in the debian defaults) and let debian install the php-7.4 versions.

I am happy to prepare a PR using either stragegy based on feedback. I think the second option is the cleanest but the first option is more backward compatible.

MegaphoneJon:
+1 for removing sury, given last week's dev-digest saying we're about to start pushing people toward 7.4.

seamuslee001:
resolved via #668

PR #668, opened 2022-01-11
Install php7.4 instead of php7.3 for various installations

seamuslee001:
ping @jmcclelland @MegaphoneJon @homotechsual @totten


MegaphoneJon:
I confirmed that all Debian/Ubuntu versions are using Ondrej Sury's apt repo. Given that, I don't see any reason why we shouldn't merge.

jmcclelland:
Thanks @seamuslee001 - this resolves #663.

PR #669, opened 2022-01-17
civilint - Enable scanning of CSS

totten:
Overview
----------

@eileenmcnaughton pointed out that [the phpcs ruleset has some coverage for CSS](https://github.com/civicrm/civicrm-core/pull/22529).

This updates `civilint` to run `phpcs` against `*.css` files.

Before
-------

* `*.php` => check with `php -l` and `phpcs`
* `*.js` => check with `jshint`

After
------

* `*.php` => check with `php -l` and `phpcs`
* `*.css` => check with `phpcs`
* `*.js` => check with `jshint`

Comments
-----------

@eileenmcnaughton @seamuslee001 @colemanw - What I'm more ambivalent about is how to activate enforcement. Right now, we always apply the same `civilint` interpretation to all PRs (eg for 5.45-stable, 5.46-rc, and 5.47-dev). Perhaps we merge the cleanup (ie https://github.com/civicrm/civicrm-core/pull/22529) into `5.46-rc` and `5.47-dev`; then, in +1 month, we merge this PR start enforcing CSS style?

### DIFF
--- a/bin/civilint
+++ b/bin/civilint
@@ -36,16 +36,17 @@ function git_default_files() {
   git diff --name-only --cached
 }
 
-## Split files into categories (PHP-files, JS-files, skip-files).
-## usage: cat file-list.txt | split_files phpfiles.txt jsfiles.txt skipfiles.txt
+## Split files into categories (PHP-files, JS-files, CSS-files, skip-files).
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
@totten it's actually mostly OK because civilint only affects changed files - the chances of us changing these files in an older version is trivially low - and it's also pretty trivial to fix up the PR to correct any offending lines if we really really really need to

PR #670, opened 2022-02-05
Use pre 2022-05 package version to permit installation of php8.1 for …

seamuslee001:
…edge profile and use an updated runkit7 version

ping @totten @eileenmcnaughton @demeritcowboy 

This built for me locally

### DIFF
--- a/nix/pkgs/php81/php.ini
+++ b/nix/pkgs/php81/php.ini
@@ -7,14 +7,13 @@ date.timezone = America/Los_Angeles
 max_execution_time = -1
 memory_limit = 256M
 
-xdebug.mode=debug
 xdebug.start_with_request=yes
 xdebug.remote_handler=dbgp
 xdebug.client_host=localhost
 xdebug.client_port=9000
 xdebug.log_level=0
 xdebug.discover_client_host=1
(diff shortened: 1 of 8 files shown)


nix/pkgs/php81/php.ini
```diff
@@ -0,0 +1,29 @@
...
+memory_limit = 256M
+
+xdebug.mode=debug
+xdebug.start_with_request=yes
+xdebug.remote_handler=dbgp
+xdebug.client_host=localhost
+xdebug.client_port=9000
+xdebug.log_level=0
+xdebug.discover_client_host=1
+xdebug.mode=develop
```
demeritcowboy, line 17: minor note - should either combine with line 10 (`xdebug.mode=debug,develop`) or remove line 10. If this is mostly for unit tests they'll run faster without "debug" and I don't think it's needed for those runs(?)

totten:
I did a bit of light `r-run` on this. Confirmed that it builds php81.

There were definitely a lot of warnings when playing around with CLI commands and D7. But I suppose that's good thing, since the whole point of `edge` is to bring up those kinds of issues...

PR #680, opened 2022-03-23
WMFF: set up directories for audit processing

ejegg:
TODO on the docker side - mount archive of logrotated files at the
directory indicated by wmf_audit_log_archive_dir.

### DIFF
--- a/app/config/wmff/install.sh
+++ b/app/config/wmff/install.sh
@@ -115,10 +115,6 @@ if [ -e "$WMF_ROLES_FILE" ]; then
   drush -y user-add-role "Engage Direct Mail" "engage"
 
 fi
-# These are used by thank you emails, eoy emails, can be removed when we
-# are no longer using the drupal emails.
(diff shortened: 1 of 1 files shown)


totten:
@ejegg You mention a `TODO` above. Does that mean we should continue holding on merge? Or is that just an aside about future topic?

ejegg:
Hi @totten, that's just a reminder to myself to look into a further bit of setup in a different repo. This patch is all set to merge for the wmff config here.

totten:
@ejegg Cool. Merged.

Depending on the environments where this runs, you might at some point get an error here (if the folder already exists):

```
mkdir /var/spool/prometheus/
```

This may depend on the specific flavor of `mkdir`.

A soft+portable variant would be either:

```
cvutil_mkdir /var/spool/prometheus/
# or
[ ! -d /var/spool/prometheus/ ] && mkdir /var/spool/prometheus/

## The second one runs almost anywhere, but it's more verbose - esp when doing multiple dirs
```

PR #682, opened 2022-03-28
civibuild restore - Add option to skip the test DB

totten:
Before:

* `civibuild restore` defaults to restoring all 3x DBs (`cms`, `civi`, `test`).
* You can opt-out of `cms` or `civi`.

After:

* `civibuild restore` defaults to restoring all 3x DBs (`cms`, `civi`, `test`).
* You can opt-out of `cms`, `civi`, or `test`

### DIFF
--- a/src/civibuild.parse.sh
+++ b/src/civibuild.parse.sh
@@ -236,6 +236,10 @@ function civibuild_parse() {
         CMS_SQL_SKIP=1
         ;;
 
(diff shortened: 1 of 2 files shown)


totten:
Note: Implementation of the option was already started before (via `TEST_SQL_SKIP`). This just needed to declare the CLI flag..

PR #694, opened 2022-05-05
(civibuild.lib.sh) backdrop_user - Don't need to Civi to add CMS user

totten:
Before: `backdrop-empty` build-type stopped working, because it needs to create
a backdrop user (`backdrop_user ...`), but the implementation (`cv ev`) relies
on booting Civi-first.

After: The implementation is basically the same, but it does need to boot
Civi (ie when adding a CMS user, it only boots CMS).


totten:
Manually deployed on test-1, and hydra-backdrop worked with this.

Issue #679, opened 2022-03-17
`civibuild reinstall` should delete afforms


MegaphoneJon:
Running `civibuild reinstall` should give a fresh install of CiviCRM.  However, it doesn't currently delete the contents of the `ang` folder so afforms created prior to reinstall persist.

colemanw:
I'm on the fence about this one. Clearing out the database is expected, but Afforms are not stored in the db.

PR #697, opened 2022-05-14
joomla-demo - Switch from git to tarball

totten:
This generally reframes what "joomla-demo" tries to offer:

* __Before__: Try to provide parity with Drupal dev builds.  Install from git, with any given version of CiviCRM and Joomla.  This is what you want for developing and testing patches on civicrm-core and civicrm-joomla, but it's complex and always had trouble.

* __After__: Install from tarball.  Only install limited versions of CiviCRM and Joomla (*ie the nightly tarball*).  This is sufficient for demo sites. It's also sufficient for developing extensions and testing cv.  (*It might also be sufficient for E2E tests, but probably with some extra patches.*)

This patch is generally hacky (sacrifices some workflows+features that are supported on other buildtypes; uses work-around for an installer bug). But...  within its narrower scope, it should be more workable.

Some things need more tuning:

* Setup permissions for the demo user
* Add some introductory content on the home page
* Configure CiviCRM with some demo data

See also: https://lab.civicrm.org/infra/ops/-/issues/982


ufundo:
hi @totten - was just wondering the rationale for dropping this?

I remember @aydun working on Joomla buildkit support last time we were in Ashbourne, maybe related?

totten:
In reverse order...

> (@ufundo) I remember @aydun working on Joomla buildkit support last time we were in Ashbourne, maybe related?

This appears to be an older PR (circa 2022; pre-dating @aydun's #882 from 2024). #882 was autoclosed while awaiting revisions -- but I resubmitted it (with some revisions) as #894 - and that was merged into `master` a while ago.

> was just wondering the rationale for dropping this?

Well, mostly, it autoclosed (awaiting revisions)... But I think maybe the real question is about the substantive difference between #679 (which uses the `joomla-console` tool) and the newer 882/894/master (which use `_joomla_download()`, `joomla_cli()`, and the `$CMS_ROOT/cli/joomla.php`).

On paper, `joomla-console` sounded good -- like the Joomla analog of `drush`. But it was in here for a few years, and there were periodic headaches -- I don't think it's as easy to work with as `drush`. (Part of that might be design; part of it might be the size of audience, #maintainers, business-model, etc.) IMHO, the approach that Aiden worked-out (now in `master`) is lighter/more maintainable.

PR #698, opened 2022-05-19
Bump up MySQL version for `old`, `min`, and `max` profiles

totten:
| ... | old | min | dfl | max | edge |
| --| --| --| --| --| --|
| Was | mysql55 | mysql56 | mysql57 | mysql57 | mysql80 |
| Now | __mysql56__ | __mysql57__ | mysql57 | ~~mysql57~~ __mysql80__ (*edited*) | mysql80 |

Aside @seamuslee001 - Shouldn't we also bump up `max=>mysql80`? I know `edge` has been failing for a while, but I thought that was mostly about php81? I use `mysql80` day-to-day -- you'd agree it's in good shape, right?

### DIFF
--- a/nix/profiles/max/default.nix
+++ b/nix/profiles/max/default.nix
@@ -13,7 +13,7 @@ in (import ../base/default.nix) ++ (import ../mgmt/default.nix) ++ [
     dists.v2111.apacheHttpd
     dists.v1809.mailcatcher
     dists.default.memcached
-    dists.default.mysql57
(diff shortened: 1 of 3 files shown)


seamuslee001:
@totten yes I would agree with that

totten:
Also, docs are quite supportive: https://docs.civicrm.org/installation/en/latest/general/requirements/#mysql-8

OK, pushed update. Editig description.

PR #700, opened 2022-06-06
Type drupal9-demo installs an unsupported version by default

demeritcowboy:
So use a non-fixed default.

Also for newer versions don't do the pear-exception thing since otherwise it fails.

### DIFF
--- a/app/config/drupal9-clean/download.sh
+++ b/app/config/drupal9-clean/download.sh
@@ -5,7 +5,7 @@
 ###############################################################################
 
 #[ -z "$CMS_VERSION" ] && CMS_VERSION=8.7.x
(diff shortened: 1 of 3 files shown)


totten:
Did some basic `r-run`. Confirmed this does pull in newer version of D9.

Changing the notation of `CMS_VERSION` could potentially impact other things, so I did a quick `grep` - but only found one place. And that had another issue. Split off an update #701.

On a policy level, I think this is fine. This really just shifts the default for local-testing. The CI job https://test.civicrm.org/job/CiviCRM-D8-Matrix/ is configured to set some alternative values of `CMS_VERSION`/`--cms-ver`, so there's still some visibility on multiple 9.x's.

seamuslee001:
demo site seems to have built fine with this patch

PR #701, opened 2022-06-06
D8+ - Fix configuration of file_private_path

totten:
Overview
----------

1. Enable `file_private_path` on D9/D10 (same way as D8)
2. Ignore composer version-constraint chars

(Off-shoot from reviewing from @demeritcowboy's #700.)

Before
-------

On Drupal 9:

```
$ drush ev 'echo Drupal\Core\Site\Settings::get("file_private_path");'
(blank - no value)
```

After
-----

On Drupal 9:

```
$ drush ev 'echo Drupal\Core\Site\Settings::get("file_private_path");'
/Users/totten/bknix/build/drupal9-clean/.civibuild/private/default
```

Comment
----------

D7 and D8 configurations use different techniques (`drush vset` vs `$settings`) for setting `file_private_path`. This is because `drush vset` stopped working in D8 (*at least, drush8 + d8/d9 doesn't support it -- and`$settings` is the documented way to configure on D8+*).

```
$ drush vset file_private_path /Users/totten/bknix/build/drupal9-clean/.civibuild/private/default
Command variable-set requires Drupal core version 6 or 7 to run.                                                                                                                           [error]
The drush command 'vset file_private_path /Users/totten/bknix/build/drupal9-clean/.civibuild/private/default' could not be executed.                                                       [error]
```

### DIFF
--- a/app/drupal.settings.d/pre.d/100-file_private_path-d8.php
+++ b/app/drupal.settings.d/pre.d/100-file_private_path-d8.php
@@ -1,7 +1,8 @@
 <?php
 
 global $civibuild, $settings;
-// Example CMS_VERSIONs: '7', '7.23', '7.x', '8', '8.8', '8.x'. Note that the '8.x' case would mess up `version_compare()`.
-if (isset($civibuild['CMS_VERSION'][0]) && empty($settings['file_private_path']) && $civibuild['CMS_VERSION'][0] === '8') {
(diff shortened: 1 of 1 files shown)


demeritcowboy:
Is privacy still a thing? (grin)
Codewise +1. I'll give it a quick run.

demeritcowboy:
Runs ok:

And at admin/config/media/file-system
```
Private file system path
/usr/local/buildkit/app/private/drupal9-demo/default 
```

PR #703, opened 2022-06-21
Update WP CLI to a version that supports php8.1

seamuslee001:
@totten @kcristiano @demeritcowboy


kcristiano:
Thank you @seamuslee001 This looks good to me.

PR #704, opened 2022-06-21
apache: disable TraceEnable

mlutfy:
ping @totten


totten:
Looks good. Tried it out locally, and it blocks `TRACE` - which AFAIK isn't needed for any extant configuration.

PR #707, opened 2022-07-07
standalone: update git repo URL

mlutfy:


mlutfy:
This is really trivial, and only affects standalone, so I took the liberty to self-merge.

PR #709, opened 2022-07-13
dev/core#2968 - Add some file layouts using D7 and different symlinks

totten:
The issue here is that implementers may use different file-structures with
various permutations of symlinks.  We'll define some specific
file-structures/symlink arrangements, with then add these configurations to
CiviCRM-E2E-Matrix.

https://test.civicrm.org/job/CiviCRM-E2E-Matrix/

### DIFF
--- a/app/config/drupal-sym1/download.sh
+++ b/app/config/drupal-sym1/download.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+## download.sh -- Download Drupal and CiviCRM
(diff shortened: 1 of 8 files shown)


totten:
(Note: I'm going to cross-link from a `civicrm-core` PR to talk about reviewing/goals.)

totten:
Contributes to https://github.com/civicrm/civicrm-core/pull/23911

PR #712, opened 2022-07-25
Remove legacy SOAP recommendation

eileenmcnaughton:
Change-Id: I76ae1d05981c2efda0c8398593ad5938cbbcae97


totten:
Yup, that makes sense. 👍 

We'll probably still install the `soap` extension on CI boxes for some time, but it should be safe to let the recommendation fade out.

PR #713, opened 2022-07-25
Update buildkit to our supported php version

eileenmcnaughton:
People (if any) who want buildkit to support a php version earlier than the one we are considering dropping should just grab an earlier buildkit


totten:
Agreed. php72 has been the min for quite ~18 months (and it's been compatible for much longer). A buildkit env running php72 should still be able to do a good deal of research/bisection/etc.

Issue #699, opened 2022-05-30
CiviBuild with older versions of CiviCRM fails


jaapjansma:
When running a command like `civibuild create civi537 --type drupal-demo --civi-ver 5.37.2` fails with 

```bash

Compiling additional files (For full details, use verbose "-v" mode.)
Compile: Generate CCL wrapper functions

In Event.php line 49:
                                                                                                                                                                                                                                           
  [TypeError]                                                                                                                                                                                                                              
  Argument 1 passed to Composer\EventDispatcher\Event::__construct() must be of the type string, null given, called in /buildkit/build/aivl/web/sites/all/modules/civicrm/vendor/civicrm/composer-compile-plugin/src/Event/CompileTaskEve  
  nt.php on line 55                                                                                                                                                                                                                        
                                                                                                                                                                                                                                           

Exception trace:
  at phar:///buildkit/bin/composer/src/Composer/EventDispatcher/Event.php:49
 Composer\EventDispatcher\Event->__construct() at /buildkit/build/aivl/web/sites/all/modules/civicrm/vendor/civicrm/composer-compile-plugin/src/Event/CompileTaskEvent.php:55
 Civi\CompilePlugin\Event\CompileTaskEvent->__construct() at /buildkit/build/aivl/web/sites/all/modules/civicrm/vendor/civicrm/composer-compile-plugin/src/TaskRunner.php:205
 Civi\CompilePlugin\TaskRunner->runTask() at /buildkit/build/aivl/web/sites/all/modules/civicrm/vendor/civicrm/composer-compile-plugin/src/TaskRunner.php:169
 Civi\CompilePlugin\TaskRunner->run() at /buildkit/build/aivl/web/sites/all/modules/civicrm/vendor/civicrm/composer-compile-plugin/src/TaskRunner.php:74
 Civi\CompilePlugin\TaskRunner->runDefault() at /buildkit/build/aivl/web/sites/all/modules/civicrm/vendor/civicrm/composer-compile-plugin/src/Command/CompileCommand.php:75
 Civi\CompilePlugin\Command\CompileCommand->execute() at phar:///buildkit/bin/composer/vendor/symfony/console/Command/Command.php:298
 Symfony\Component\Console\Command\Command->run() at phar:///buildkit/bin/composer/vendor/symfony/console/Application.php:1015
 Symfony\Component\Console\Application->doRunCommand() at phar:///buildkit/bin/composer/vendor/symfony/console/Application.php:299
[...]

PR #715, opened 2022-08-03
civi-download-tools - Pin composer version (#699)

totten:
ping @jaapjansma @seamuslee001 

See also: #699 and discussion on https://github.com/civicrm/civicrm-buildkit/commit/177c3ace297db792ef4b417f521c03292b1adce8


seamuslee001:
@totten just wondering if we need to consider if there is `/usr/local/bin/composer` or not

totten:
@seamuslee001 Well...  wouldn't it work the same as `drush` or `cv` or `phpunit`? It's generally true that any tool included buildkit could also be installed somewhere else. The `PATH` determines what version will be used.

PR #716, opened 2022-08-04
Create extern directory before using in install_composer

ejegg:
install_composer expects the directory to exist already, so move
the directory creation up a ways.


eileenmcnaughton:
Merging as a couple of people at WMF have been testing this so their review seems good (plus seems fairly safe)

totten:
Oops. Good catch! 👍

PR #717, opened 2022-08-11
fetch-universe - Multiple updates

totten:
* Add option `--dry-run`
* Add option `--feeds=<CSV>` (eg `--feeds=static,extdir,lab`) to only fetch certain feeds
* Add some WP plugins from svn (*These are listed in `civicrm.org/extensions` but don't appear in the extension or gitlab feeds.*)
* Move the static feed to a separate file (`src/universe.php`)


src/universe.php
```diff
@@ -0,0 +1,159 @@
...
+    'civix' => array(
+      'git_url' => 'https://github.com/totten/civix',
+      'type' => 'tools',
+    ),
+    'coworker' => array(
+      'git_url' => 'https://lab.civicrm.org/dev/coworker',
+      'type' => 'tools',
+    ),
+    'cf-civicrm' => array(
+      'git_url' => 'https://github.com/mecachisenros/cf-civicrm',
```
kcristiano, line 68: @totten This repo is abandoned.  This repo - https://github.com/WPCV/cf-civicrm  is the fork with patches to work with CiviCRM 5.41+

kcristiano (commented):
If we are doing cleanup updating the git repo for cf-civicrm would be a good idea.

src/universe.php
kcristiano: I missed that this was merged.   I'll try and send up a PR when I get a moment

totten: @kcristiano Yup, that would be good. Maybe those other repos from `WPCV` should be added as well?

PR #719, opened 2022-08-13
fix extensions not installing on wp-demo.  Update universe

kcristiano:
add URLs for cf-civicrm and woo-civicrm

Before:

Extension failures on wp-demo buildkit and universe pointing to an out of data repo.

After:

Extensions install properly on wp-demo  and universe updated.


totten:
Cheers!

PR #727, opened 2022-09-16
nix - Update to v22.05. Adjust XDebug, MailHog, et al.

totten:
This is a general bundle of updates from `nixpkgs`. The impetus is to improve compatibility with ARM64 (aka `aarch64` aka "Apple M1/M2"). I've been using some hacks on a separate branch for a few months, and I'd like get these normalized.

Summary of updates:

* __Platform Version__: Change the `default` flavor of `nixpkgs` from `21.11` to `22.05`. Switch several more packages to track the `default`. Update references for `22.05-pre` (*we can use the final release instead of the pre-release*).
* __PHP Debugging__: Consistently use XDebug v3 on all the main profiles (`min`, `dfl`, `max`). Rearrange/simplify XDebug configuration. (*More discussion below.*)
* __Email Simulation__: Replace `mailcatcher` with `MailHog`. (`mailcatcher` has build problems in v22.05/ARM64.)
* __ARM64 Version Substitutions__: Some old versions of PHP/MySQL don't builds on ARM64. Use alternatives. (*Ex: For `min`, substitute `php72` with `php74`.*)

Deployment notes:

* For local dev: Close any shells/IDEs that rely on nix. Then run `install-developer.sh`. Start shells/IDEs again.
* For CI: Use the `reset-ci.sh` procedure. (Or reboot on nodes that auto-update.)
* Before merging: I'd like to run Civi's test-suite on a temp-node.

The XDebug changes could potentially affect development/debugging, so let me break it down a bit:

* __Version__: Some builds (php72, php73) had XDebug v2, and others (php74, php80, php81) had XDebug v3. Now, they all use the same version (3.1).
* __Default Port__: In v2, XDebug's default port was 9000. In v3, it's 9003.
* __Configuration Reunification__: Several configuration directives changed between XDebug v2=>v3. When we had mixed versions, the configuration was split/duplicated. Now, it's reunified (ie php72-php81 all use the same XDebug config).
* __Environment Variables__: With all builds in XDebug v3, we can standardize on using environment-variables for common XDebug options. This makes it easier to enable/disable on-the-fly. It's also easier to differentiate options by environment. For example:
    * When a developer manually calls `loco run`, it will set `XDEBUG_MODE=develop,debug`. (*This can be changed in `loco.yml`.*)
    * When a test-server calls `loco run ...`, it will set `XDEBUG_MODE=off`. (*This can be changed in `jenkins-max.yml`, etal.*)
    * If you need an alternate/local configuration, it's faster to update the env-var than to update `php.ini`. (To update nix's `php.ini`, you have to reinstall the packages.)

I've done a fair amount of `r-run` for XDebug with PHPStorm 2021.x on Ubuntu and MacOS, running many permutations of PHP version (7.2, 7.4, 8.0, 8.1) and command-runners (eg manual CLI; manual web-request; IDE-based phpunit debugging). This configuration should be working on all of them.


totten:
> Before merging: I'd like to run Civi's test-suite on a temp-node.

That also ran properly. There were test failures --  but they matched the pre-existing failures.

PR #728, opened 2022-09-23
package.json - Mark karma/phantom as optional

totten:
npm doesn't install phantomjs on linux/arm64. It's only needed for a small part of the test suite. Don't block other uses.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5056 +1,56 @@
 {
   "name": "civicrm",
   "version": "0.1.0",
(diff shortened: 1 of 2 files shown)


totten:
I haven't used `optionalDependencies` before. Did a quick bit of `r-run` (checking out the branch on other systems; reinstalling; deleting `node_modules`; reinstall again; etc), and it seemed to do the right thing (ie installing on linux/x86 and macos/arm64 but skipping the unavailable linux/arm64).

PR #730, opened 2022-10-09
Update box version to latest in the 3.x branch

seamuslee001:
ping @totten


seamuslee001:
Note that this is the same box version as in cv https://github.com/civicrm/cv/blob/master/composer.json#L39

totten:
FWIW, some of the PHAR builders have switched to using their own copy of `box`. It's worth a look to see if any of https://test.civicrm.org/view/Tools/ still rely on the version from `box` from buildkit.

PR #731, opened 2022-10-22
Update DFL and Max Profiles to use PHP8.1

seamuslee001:
@totten @eileenmcnaughton


eileenmcnaughton:
Looks good to me - merging as this was discussed & agreed on Friday

Side note - I learnt that the 'dfl' profile is actually an abbreviation for something at the dev training (I can't recall what though 'cos it wasn't something you would think of from 'dfl' - but for a brief period I did know).

PR #734, opened 2022-10-28
bknix-edge - Add PHP 8.2 and XDebug 3.2

totten:
Upstream `nixpkgs` doesn't officially have PHP 8.2 yet. That seems fair -- given that it's still only in RC.

But it wasn't [too hard to get a preview build](https://github.com/totten/nixpkgs/commit/3673729e7edc160135299a87e781981351fb8c2d). I based it on nixpkgs 22.05 (so it should mostly use the same low-level dependencies as the other PHP builds).

There's still some PECL extensions that need to shake-out. For the moment, I backported XDebug 3.2.0rc1 (which adds support for PHP 8.2), and temporarily skipped some less-used PECLs (`apcu`, `tidy`, `runkit7_4`).

### DIFF
--- a/nix/pins/22.05-php82.nix
+++ b/nix/pins/22.05-php82.nix
@@ -2,6 +2,6 @@
  * v22.05 with unofficial php82 backport (`totten/nixpkgs @ php82-v2205`)
  */
 fetchTarball {
-  url = "https://github.com/totten/nixpkgs/archive/3673729e7edc160135299a87e781981351fb8c2d.tar.gz";
-  sha256 = "1cclziiggh4s680kfz2290qzpwvn7fr0p8qlhwz9adyilsj225q7";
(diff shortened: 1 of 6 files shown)


seamuslee001 (approved):

totten:
@seamuslee001 Any reasons to hold off  `bknix-edge=>php82` ?

My inclination is to merge this in a few days -- and do so in tandem with a `bknix-min=>php73` update (#741) . Then all the binary/image updates can go at the same time.

seamuslee001:
@totten apart from it possibly blowing up the Jenkins log I don't think anything is really holding this back

totten:
PHP 8.2.0-final and XDebug 3.2.0-final went out a few days. 

Rebased and updated to latest php+xdebug.

PR #735, opened 2022-11-03
drupal-clean, drupal-demo - Switch to civicrm_install_cv

totten:
Resubmission of #673 / #661.

I've started a local run of `civi-test-run all` on a `drupal-clean` site to make sure it still runs the full suite.

### DIFF
--- a/app/config/drupal-clean/download.sh
+++ b/app/config/drupal-clean/download.sh
@@ -21,6 +21,9 @@ pushd "$WEB_ROOT/web"
 
     extract-url --cache-ttl 172800 civicrm=http://download.civicrm.org/civicrm-l10n-core/archives/civicrm-l10n-daily.tar.gz
     ## or https://raw.github.com/civicrm/l10n/master/po/fr_CA/civicrm.mo => civicrm/l10n/fr_CA/LC_MESSAGES/
+    pushd civicrm
+      composer install
+    popd
   popd
 
 popd
--- a/app/config/drupal-clean/install.sh
+++ b/app/config/drupal-clean/install.sh
@@ -38,6 +38,11 @@ civicrm_install_cv
 
 ###############################################################################
 ## Extra configuration
+pushd "$CIVI_CORE" >> /dev/null
+  ## Generating `civicrm.config.php` is necessary for `extern/*.php` and its E2E tests
+  ./bin/setup.sh -g
+popd >> /dev/null
+
 pushd "${CMS_ROOT}/sites/${DRUPAL_SITE_DIR}" >> /dev/null
 
   drush -y updatedb
--- a/app/config/drupal-clean/install.sh
+++ b/app/config/drupal-clean/install.sh
@@ -34,13 +34,13 @@ if [[ "$CIVI_VERSION" =~ ^4.[0123456](\.([0-9]|alpha|beta)+)?$ ]] ; then
   CIVI_EXT_URL="${CMS_URL}/sites/${DRUPAL_SITE_DIR}/ext"
 fi
 
-pushd "$CIVI_CORE" >> /dev/null
-  ## (1) Generating `civicrm.config.php` is necessary for `extern/*.php` and its E2E tests
-  ## (2) If you've switched branches and triggered `reinstall`, then you need to refresh composer deps/autoloader before installing
-  ./bin/setup.sh -Dg
-popd >> /dev/null
+# If you've switched branches and triggered `reinstall`, then you need to refresh composer deps/autoloader before installing
+(cd "$CIVI_CORE" && composer install)
 
 civicrm_install_cv
+
+## Generating `civicrm.config.php` is necessary for `extern/*.php` and its E2E tests
+(cd "$CIVI_CORE" && ./bin/setup.sh -g)
 
 ###############################################################################
 ## Extra configuration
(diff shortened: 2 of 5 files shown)


totten:
On new local build with `drupal-clean`, I tried `civi-test-run` for `phpunit-e2e`, `phpunit-core-exts`, and `mixin`. Almost everything was passing... except for `search_kit`'s `headless` suite.

@colemanw Any chance you could checkout this branch and try something like:

```bash
civibuild create tmp --type drupal-clean 
civi-test-run -b tmp -j /tmp/junit phpunit-searchkit
```

It should generate something like this: https://gist.github.com/totten/f2efd3f45dbf5cea55d10bc0f7fece3f

colemanw:
@totten how do I checkout a branch *before* installing a build?

totten:
There's a secondary issue which I don't fully understand yet.

```
Installing dmastertest_qmzfj schema

Fatal error: Uncaught Error: Class 'Civi\Api4\SearchSegment' not found in /Users/totten/bknix/build/dmaster/web/sites/all/modules/civicrm/ext/search_kit/Civi/Api4/Service/Spec/Provider/SearchSegmentExtraFieldProvider.php on line 52

Error: Class 'Civi\Api4\SearchSegment' not found in /Users/totten/bknix/build/dmaster/web/sites/all/modules/civicrm/ext/search_kit/Civi/Api4/Service/Spec/Provider/SearchSegmentExtraFieldProvider.php on line 52
```

Context:

* It regards PHPUnit tests in extensions.
* It appears in 5.51 but not 5.56 or master.
* It appears when SK is enabled in the base SQL snapshot (ie when `civibuild restore` gives you a system with SK active).  This PR makes it appear more often -- before, it only appeared with `drupal-demo` and `wp-dem`; with this, it will also appears in `drupal-clean` (which is used by most CI test-runs).
* It manifests with the test-suites in `mosaico`, `extendedreport`, and in newly-generated extensions.
    ```
    cd civix
    export CIVIX_WORKSPACE=/path/to/civicrm/ext/civixtest
    ./scripts/make-snapshots.sh --src --test --keep --version HEAD entity3
    ```

* Related: https://test.civicrm.org/job/CiviCRM-Civix-Matrix/ (note the 5.51 column - passing on old/non-cv flavor of drupal-clean which lacks SK; failing on the old drupal-demo/wp-demo flavors which include SK).

After running tests more times than I care to recall... it seems this issue has gone away circa 5.52+ -- so maybe it's not critical from core pov.

(The prior comment though -- regarding SK's test-suite -- is still problematic and would affect `master`.)

totten:
> @colemanw how do I checkout a branch before installing a build?

Not certain the question - here a few answers:

* For loading 735, that's just needs to be loaded onto the `buildkit.git` repo first.
* If you want to create a civibuild that includes a draft PR, then:
    * The easier way IMHO:
        ```
        civibuild create tmp --type drupal-clean --patch GITHUB_PR_URL
        ```
    * The more manual way is to split the `create` step into smaller `download` and install steps, eg 
        ```bash
        civibuild download tmp --type drupal-clean
        # Do any code manipulation you want
        # Then continue with installation:
        civibuild install tmp
        ```

colemanw:
@totten whether you're still seeing it in 5.56 or not, here is a safeguard which IMO should be merged in regardless: #25114
will prevent those errors popping up again.

totten:
OK, cool, applying 25114 on 5.51 does fix the `SearchSegment` errors (at least, for the civix `make-snapshot` use-case). I also tried applying 24977, but it didn't seem to make much difference -- probably because it's focused on AJAX+Smarty entry-points (which shouldn't be in the callpath for an empty/skeletal test). 🤷 

I also did a new run of `phpunit-searchkit` just to make sure these patches didn't influence that. This turned out to be a rabbit hole that can only be endured by dedicated rabbits. Following my own suggested steps above, I could not initially reproduce. I banged my head against different revisions of `master`, different versions of PHP, and enough rebuilds to make your vision go blurry -- just couldn't get it to reproduce. But I had seen it several times before, and it made no sense for the error to go away. Why couldn't I see it anymore?

Well... Finally found a thread to pull. Those failures appear to be an interaction between the various test-suites. So it recurs if you use `drupal_clean_install_cv` _and_ run _all_ the core-ext tests; thereafter, it will recur again one more time. But if _only_ run the SK suite in isolation, then it passes.

```bash
## Flow 1
civi-test-run -b tmp -j /tmp/junit-solo phpunit-searchkit  ## Pass
civi-test-run -b tmp -j /tmp/junit-solo phpunit-searchkit  ## Pass
civi-test-run -b tmp -j /tmp/junit-solo phpunit-searchkit  ## Pass

## Flow 2
civi-test-run -b tmp -j /tmp/junit phpunit-core-exts       ## 10 failures
civi-test-run -b tmp -j /tmp/junit-solo phpunit-searchkit  ## 10 failures
civi-test-run -b tmp -j /tmp/junit-solo phpunit-searchkit  ## Pass
```

totten:
OK, the interaction involves civiimport<=>search_kit. This reproduces the problem:

```
git checkout drupal_clean_install_cv

civibuild create tmp --type drupal-clean
cd build/tmp/web/sites/all/modules/civicrm/ext/

(civibuild restore && cd civiimport && phpunit8 --group headless )
(civibuild restore && cd search_kit && phpunit8 --group headless )
## ^ Produces 10 failures, like https://gist.github.com/totten/f2efd3f45dbf5cea55d10bc0f7fece3f
```

totten:
Throwing in a `cv flush` does fix it, as in:

```bash
(civibuild restore && cv flush && cd civiimport && phpunit8 --group headless )
(civibuild restore && cv flush && cd search_kit && phpunit8 --group headless )
```

Interestingly, deleting `sites/default/files/civicrm/templates_c` doesn't help. (The folder isn't even populated after the test.) So there must be some other data-structure that leaks.

PR #739, opened 2022-11-25
bin/civi-download-tools - Switch to composer 2.2.18

totten:
* Before #715, the `composer` version was floating. 
* With #715, the `composer` version was pinned to whatever I had on my workstation at the time (2.3.10). However, I didn't do any testing to determine compatibility.
* Coleman did some testing and found 2.3.10 problematic in the range of Civi 5.30's-5.40's. In local testing, composer 2.2.18 seems to work with `master` and with some of those older versions.


colemanw:
Just tested and that version worked much better, allowed building sites < `4.48`. :+1:

PR #752, opened 2023-01-03
civibuild - When using "restore", the "-f" option will flush (forcibly resetting caches)

totten:
Anecdotally, when improvising test routines to isolate problems, I will often choose between:

* `civibuild restore dmaster` (which preserves caches from the snapshot)
* `civibuild restore dmaster && cv flush` (which resets caches)

(Both seem useful to me, and I don't think there's a quick rule for when to choose each.)

This PR adds a shorthand for the latter (`civibuild restore dmaster -f`). 

@eileenmcnaughton You could use this to mitigate https://lab.civicrm.org/dev/core/-/issues/4054 -- i.e. when using Redis, just include the `-f` flag.

### DIFF
--- a/src/civibuild.parse.sh
+++ b/src/civibuild.parse.sh
@@ -127,7 +127,7 @@ function civibuild_parse() {
     shift
 
     case "$OPTION" in
-      -h|--help|-?)
+      -h|--help|-\?)
         civibuild_app_usage
         ;;
 
@@ -196,7 +196,7 @@ function civibuild_parse() {
         shift
         ;;
 
-      --force)
+      -f|--force)
         FORCE_DOWNLOAD=1
         FORCE_INSTALL=1
         ;;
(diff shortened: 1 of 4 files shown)


eileenmcnaughton:
@totten that looks promising - I only ever use `civibuild create` not `civibuild restore` - but assuming it works with `civibuild create` too  we could add a bash alias for our docker to always use `-f` with `civibuild create` (or make everyone remember it  - which may or may not be more challenging)

totten:
Aaah, thanks for clarifying @eileenmcnaughton. I hadn't even considered `civibuild create`!

I don't think this PR will help for that. (It doesn't hurt either, though.) The `--force`/`-f` option has an existing meaning on `civibuild create` and `civibuild install`.

* `civibuild create -f` -- Download and install entire build, overwriting existing code+db
* `civibuild install -f` (aka `civibuild reinstall`) -- Run the Civi/CMS/DB install routines, overwriting existing db
* (See also: https://docs.civicrm.org/dev/en/latest/tools/civibuild/#rebuild)

For what you're describing, I feel like maybe we need a step in Civi's installation process -- to clear out Redis caches before they're ever used. Alternatively, we could try to re-generate a different `CIVICRM_DB_CACHE_PREFIX` every time it runs the installer?

eileenmcnaughton:
@totten my preference is the first - clear out Redis caches before they are used.

Issue #751, opened 2022-12-28
[regression] Only run `drush language add` if there are languages to add


eileenmcnaughton:
My colleague is hitting an error here because `$NEW_LOCALES` is empty - a simple check before running it would prevent :-)

https://github.com/civicrm/civicrm-buildkit/commit/010a393d22b27edc84698f922fac53cbb8ce3bae#diff-7224d122e78711d88854f3fcc1c90b57ac817a458b24a2722f98852cf330b98fR1283 

He checked out an older version of buildkit & it worked fine

eileenmcnaughton:
@totten not sure why this only affects his install not mine but the drush command is newly added

ejegg:
I can reproduce this on a new install when I completely blow away my existing (wmff-flavored) drupal / civi build.

PR #754, opened 2023-01-04
Check for empty langs in drupal PO download

ejegg:
Adds the same early return in _drupalx_po_import as we are already using in _drupalx_po_download.

BEFORE: install fails when no new languages need to be added

NOW: install completes when no new languages need to be added

Fixes: #751


totten:
Looks pretty sensible to me.

PR #761, opened 2023-02-06
civi-test-run - Add support for more filters

totten:
Before: The option `--exclude-group` will pass-through to phpunit

After: The options `--group`, `--exclude-group`, and `--filter` will all pass-through to phpunit

### DIFF
--- a/bin/civi-test-run
+++ b/bin/civi-test-run
@@ -78,7 +78,7 @@ function show_help() {
 function task_phpunit() {
   SUITE="$1"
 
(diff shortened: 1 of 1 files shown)


seamuslee001:
@totten I'm guessing there isn't any idea of trying to combine filters like `--group --filter` together

totten:
Good call. There should be... I guess `TEST_FILTER` should be a bash array, eg

```bash
TEST_FILTER+=("--exclude-group" "$1")
TEST_FILTER+=("--group" "$1")
TEST_FILTER+=("--filter" "$1")
```

And then I *think* the final run-commands would be like `phpunit ${TEST_FILTER[@]}` (*but my recollection about bash arrays is foggy*).

totten:
Thanks @seamuslee001

Issue #763, opened 2023-03-01
Update transifex-cli (cannot update transifex)


mlutfy:
The transifex-cli is now a Go package, instead of Python. The old client is not supported anymore, the API calls fail, so translations cannot be updated.

I found this blogpost, but I could not figure out how to update nix:  
http://lethalman.blogspot.com/2015/02/developing-in-golang-with-nix-package.html

I poked around here:  
https://github.com/civicrm/civicrm-buildkit/tree/master/nix/pkgs/transifexClient

And as far as I understand, we need something like this:

```
{ goPackages, lib, fetchFromGitHub, six, requests }:
buildGoModule rec {
  pname = "transifex-cli";
  version = "1.6.5";

  src = fetchFromGitHub {
    owner = "transifex";
    repo = "transifex-cli";
    rev = "${version}";
    sha256 = "FIXME"; // ?
  };

  modSha256 = "FIXME"; // ?

  subPackages = [ "cmd" ]; // random copy-paste, no idea what this is for

  meta = with stdenv.lib; {
    description = "Transifex CLI";
    homepage = "https://github.com/transifex/transifex-cli";
    license = licenses.asl20;
  };
}
```

but I just get this:

```
$ nix-shell -A min 
error: undefined variable 'buildGoModule'
```

for which there are zero helpful google results.

In `pkgs/transifexClient/default.nix`, I also tried replacing the python ref with:

```
transifexClient = import ./transifexClient.nix { };
```

but without success, same error.

All this feels weird, because as far as I understand, nix wants to build from source, but with go, it's much easier to use the binaries.

totten:
@mlutfy That does have the about the right shape. A couple general thoughts:

* There's also a doc at https://nixos.org/manual/nixpkgs/stable/#sec-language-go. They tend to be more recent but also a bit thin.
* When I get confused, I often grep https://github.com/NixOS/nixpkgs/ for examples of similar packages. (*Compared to the manual and the blogs, that repo tends to be more recent+working.*)
* There is one big thing that differentiates `civicrm-buildkit` from canonical nix. The opening declaration like `{foo, bar}:` is often replaced with an explicit `import`. Noted here: https://github.com/civicrm/civicrm-buildkit/pull/764/commits/8b2258a3d432923638ac745fd127d62134acac19
* I took the example from above and did several small iterations in https://github.com/civicrm/civicrm-buildkit/pull/764. You should be able to check that out and try the `nix-shell -A min` again.

[...]

PR #764, opened 2023-03-01
Update transifex

totten:
For #763


totten:
There's a confusing thing in here:

```
$ nix-shell -A min
[nix-shell:~/bknix]$ which tx 

[nix-shell:~/bknix]$ which cli
/nix/store/hd10f50d7m8fc0bfx813q10yzzr2q719-transifex-cli-1.6.5/bin/cli
```

So it's generating a command named `cli` (which, in fairness, matches the Github project - but it's a bit nondescript). 

It probably needs some kind of extra step. Maybe an `installPhase` step to copy "cli" to "tx"? Or maybe there's an example somewhere in nixpkgs for telling `buildGoModule` to rename the binary?

totten:
Fixing the command-name basically requires adding an extra build step (e.g. `mv bin/cli bin/tx`). Unfortunately, I don't do this enough to _know_ where to put that. But... I was able to find the [implementation of buildGoModule](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/go/module.nix), spot a few key words (`$GOPATH`, `runHook`, `postBuild`), and make a lucky guess:

```
  postBuild = ''
    mv $GOPATH/bin/cli $GOPATH/bin/tx
  '';
```

Which seems to work.

But, dang, I really wish the docs described the parameters for these functions like `buildGoModule`.

mlutfy:
Wow thanks for helping with this, definitely obscure. There is a warning, but it seems to work.

I ran it on my local env, and initially had this non-fatal warning:

```
0 bgm@home:~/repositories/civicrm-buildkit((HEAD detached at totten/master-tx-165))$ nix-shell -A min 
error: file 'nixpkgs' was not found in the Nix search path (add it using $NIX_PATH or -I)

       at «string»:1:9:

            1| (import <nixpkgs> {}).bashInteractive
             |         ^
will use bash from your environment
these 3 derivations will be built:
  /nix/store/bd7976f49wq6397bwbivws71316ibl91-source.drv
  /nix/store/z3says2pi5x13c0mq0ixhzl4a1x52kkm-transifex-cli-1.6.5-go-modules.drv
  /nix/store/2zfwkhr4vigykgca6bqkrmkj490v72xz-transifex-cli-1.6.5.drv
this path will be fetched (96.73 MiB download, 428.78 MiB unpacked):
  /nix/store/jsarcxgrisc7w5znkja44lqqy4zv8cs5-go-1.17.10
copying path '/nix/store/jsarcxgrisc7w5znkja44lqqy4zv8cs5-go-1.17.10' from 'https://cache.nixos.org'...
```

and when it was done, it seemed to work OK:

```
[nix-shell:~/repositories/civicrm-buildkit]$ which tx 
/nix/store/a7dki11238px2sw17wjy6xdx0c939fk9-transifex-cli-1.6.5/bin/tx

[nix-shell:~/repositories/civicrm-buildkit]$ tx 
NAME:
   tx - A new cli application
[...]
```

The command line options seem to be the same as before, so I would be tempted to merge and to run it. It's a bit difficult to test the full process locally.

totten:
> The command line options seem to be the same as before, so I would be tempted to merge and to run it. It's a bit difficult to test the full process locally.

Agree, let's just merge. It sounds like the old client doesn't work anyway, so we might as well try it.

PR #767, opened 2023-03-09
standalone - A few fixes

totten:
* Auto-publish assets
   * Note: I personally like [symdir style](https://lab.civicrm.org/dev/civicrm-asset-plugin/#options) and opened a separate (#766) as way to encourage it without forcing it everywhere.
* Mark data folder permissions
   * @artfulrobot This expands on [this line](https://github.com/artfulrobot/civicrm-buildkit/blob/3641381ec8f1d4f1f76d52db2f2e7958af41b21a/app/config/standalone-clean/download.sh#L17) in your branch. The command `amp datadir` is a bit like `chmod a+rwx`, but it adapts to [local policy](https://github.com/amp-cli/amp/blob/master/doc/perm.md).
* Fix support for `civibuild reinstall <sitename>`
    * This requires some extra bits for removing the old configuration.

### DIFF
--- a/app/config/standalone-clean/download.sh
+++ b/app/config/standalone-clean/download.sh
@@ -7,6 +7,11 @@
 # @todo Requires that we register with packagist?
 # mkdir "$WEB_ROOT"
 # composer create-project civicrm/civicrm-standalone "$WEB_ROOT" --no-interaction
-git clone https://github.com/civicrm/civicrm-standalone $WEB_ROOT
-cd $WEB_ROOT
-composer install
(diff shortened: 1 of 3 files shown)


totten:
Updated to also respect the `--cms-ver` flag. This should work properly whenever we start tagging things in `civicrm-buildkit.git`, and (for now) it allows you to request a fork without needing any local patches. Ex:

```bash
civibuild download standalone-clean --cms-ver 'artfulrobot/artfulrobot-clean-install-paths'
```

PR #770, opened 2023-03-31
Enable civicrm_admin_ui message_admin org.civicrm.afform on demo sites

eileenmcnaughton:
What about afform_admin?


totten:
If we're adding `message_admin` and `civicrm_admin_ui`, then I'd also be +1 for `afform_admin`.

PR #773, opened 2023-04-12
Remove platform override and update drush to use upstream 8.4.12 vers…

seamuslee001:
…ion and update wp to latest wp cli

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,9 +12,6 @@
     }
   },
   "config": {
-    "platform": {
-      "php": "7.2"
-    },
     "bin-dir": "bin",
     "allow-plugins": {
       "civicrm/composer-downloads-plugin": true,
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -12,9 +12,6 @@
     }
   },
   "config": {
-    "platform": {
-      "php": "7.2"
```
totten, line 16: Is there a reason for removing (as opposed to raising it to 7.3)? When it's unlocked, it's more prone to crossed wires in `composer.lock`, no?

seamuslee001: @totten if we had to have one I would think 7.4 or 8.0 makes more sense on a development repo these days but I guess I am not aware of any specific project here that would have issue with this being removed, Generally it is better practice to not be overriding platform I think

PR #777, opened 2023-05-23
expand the Universe for WP Plugins in use

kcristiano:
Add Frequently used plugins to Universe.


totten:
Ah, great. Thanks @kcristiano.

PR #780, opened 2023-06-08
Add build profiles (`drupal10-empty` and `drupal10-clean`)

totten:
jfyi @seamuslee001 @demeritcowboy, this was a pretty straight adaptation from the d9 build scripts. The main changes:

* Switch a couple constants/defaults (`^9` => `^10`)
* Add upgraded version of drush (`composer require drush/drush`)
* Opt-out of some old hacks in `civicrm_download_composer_d8`

With this, I've been able to tentatively run some tests:

```bash
civibuild create d10 --type drupal10-clean
civi-test-run -b d10 -j /tmp/junit phpunit-e2e
civi-test-run -b d10 -j /tmp/junit phpunit-civi
### Aborted because I'm impatient, but it was runnig OK for a while
```


demeritcowboy:
Nice.

I have a drupal 10 build running nightly with a subset of core tests, but I needed to deal with phpunit9 since 8 isn't supported. Curious how you made that work since I needed to patch core tests.

This is what I'm running: https://github.com/demeritcowboy/testCI/blob/f118509e53ab2050e93e581ecf4ee5505a273d1e/.github/workflows/drupal10.yml#L64-L70

seamuslee001:
I thought there would be PHP Version issues but maybe that isn't a big thing on people's locals

totten:
> (@demeritcowboy) Curious how you made that work since I needed to patch core tests.

Well, I copied this bit from the `drupal9-clean` -- it uninstalls drupal's phpunit. Perhaps that's why?

https://github.com/civicrm/civicrm-buildkit/blob/7342083ce8982574a90ab2700c80263f55de9215/app/config/drupal10-clean/download.sh#L19-L24

> (@seamuslee001) I thought there would be PHP Version issues but maybe that isn't a big thing on people's locals

Oh, yeah, I couldn't use `min` (php73). I've been able to do with `max` (php81).

bin/civi-download-tools
```diff
@@ -22,7 +22,7 @@ LOCKFILE="$TMPDIR/civi-download-tools.lock"
 LOCKTIMEOUT=90
 HUB_VERSION="2.14.2"
 JOOMLATOOLS_CONSOLE_VERSION="v1.5.9"
-COMPOSER_VERSION="2.2.18"
+COMPOSER_VERSION="2.3.10"
```
totten, line 25: My main hesitation with merging was this -- for some D10 scenarios (e.g. installing directly from `drupal.git`), you need composer@2.3, but this breaks support for some older versions of Civi.

After testing that a bit, the breaking-point appears to be Civi 5.47 -- i.e. composer@2.3.10 can install civicrm@5.48 but not civicrm@5.47.

Discussed a bit with Seamus on MM, and we both felt that was acceptable -- it's good for us to support D10 in more CI scenarios, and you can still install the past 2 ESR's (5.51 and 5.57).

PR #784, opened 2023-06-21
civilint - Call "civistrings" and look for warnings

totten:
@mlutfy - I was reading [this comment](https://github.com/civicrm/civicrm-core/pull/26565#issuecomment-1599288709) and took a quick stab at updating civilint. In principle, this would cause the style-checks to fail if a modified file has any warnings from `civistrings`.

There are two things I'm not sure about here:

1. The specific error message would only be on the console. (Takes a little more work to bubble-up to the report in Jenkins.) But that's livable (and maybe fixable).
2. It may be a little too zealous? Like, it would enforce style-rules on pre-existing things in all branches (e.g. current patch went into 5.64-dev; so 5.63-rc and 5.62-stable still have more messy bits). And there are a few oddballs where it's intentionally deviant (e.g. `Civi/Angular/Manager.php` `getTranslatedStrings()`).

### DIFF
--- a/bin/civilint
+++ b/bin/civilint
@@ -219,6 +219,16 @@ if grep -q . "$TMP/js.txt" ; then
   fi
 fi
 
+CIVISTRINGS_EXIT=0
+if grep -q . "$TMP/js.txt" "$TMP/php.txt" ; then
+  echo "============================[ civistrings ]============================="
+  cat  "$TMP/js.txt" "$TMP/php.txt" | ( xargs $BINDIR/civistrings 2>&1 ) | grep '^\[warn' > "$TMP/civistrings-check.log"
+  if grep -q . "$TMP/civistrings-check.log" ; then
+    CIVISTRINGS_EXIT=1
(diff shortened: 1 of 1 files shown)


mlutfy:
@totten :tada:  this would be great

For (1), would it still cause the build to fail? (so that we can investigate, see the error). If yes, sounds OK to me. We can tweak later.

For (2), maybe - there are legitimate situations where someone might do `ts($name)` - not many, but in some obscure places of core or SearchKit/FB, it happens. Maybe we could have a wrapper `ts`  function that is for those situations?

totten:
For (1), yes it would fail, so it does that job. To make it fail more nicely, I've a separate issue on the backlog (https://github.com/civicrm/civistrings/issues/14).

For (2), I like that idea and opened https://github.com/civicrm/civicrm-core/pull/26595. (Still need something else to deal with the 5.63-vs-5.64 issue -- for that aspect, feels like linter needs to somehow incorporate the branch/version.)

PR #788, opened 2023-07-09
Standalone - Support for HAN-C and HAN-D

totten:
There will be two standalone profiles:

* `standalone-clean`: As today, install `civicrm-standalone.git` (and recursively load other packages like `civicrm-core`) (*HAN-C*)
* `standalone-dev`: Install `civicrm-core` directly (*HAN-D*)

### DIFF
--- a/app/config/standalone-clean/install.sh
+++ b/app/config/standalone-clean/install.sh
@@ -19,6 +19,12 @@ CIVI_SETTINGS="${WEB_ROOT}/data/civicrm.settings.php"
 CIVI_TEMPLATEC="${WEB_ROOT}/data/templates_c"
 GENCODE_CONFIG_TEMPLATE="${CMS_ROOT}/civicrm.config.php.standalone"
 
+pushd "$CIVI_CORE"
+  ./tools/standalone/bin/scaffold "$WEB_ROOT"
+  ## This may technically be a bit redundant with 'composer install' for new builds.
+  ## But for long-lived sites that have rebuilds, it's handy.
+popd
+
 civicrm_install_cv
 
 composer civicrm:publish
--- a/app/config/standalone-dev/demo-user.php
+++ b/app/config/standalone-dev/demo-user.php
@@ -1,0 +1,38 @@
+<?php
+
+// Add a demo user.
+
+if (PHP_SAPI !== 'cli') {
+  die("phpunit can only be run from command line.");
+}
+
+foreach (['DEMO_EMAIL', 'DEMO_USER', 'DEMO_PASS'] as $var) {
+  if (empty(getenv($var))) {
+    throw new \Exception("Missing: $var");
+  }
+}
+
+// Since users+contacts are stored in same DB, we can do transaction.
+CRM_Core_Transaction::create()->run(function () {
+
+  $contactID = \Civi\Api4\Contact::create(FALSE)
+    ->setValues([
+      'contact_type' => 'Individual',
+      'first_name' => 'Demo',
+      'last_name' => 'User',
+    ])
+    ->execute()->first()['id'];
+  $adminEmail = getenv('DEMO_EMAIL');
+  $params = [
+    'cms_name' => getenv('DEMO_USER'),
+    'cms_pass' => getenv('DEMO_PASS'),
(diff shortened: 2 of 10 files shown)


totten:
@artfulrobot  Following on some ideas from https://lab.civicrm.org/documentation/docs/installation/-/merge_requests/56, I played around, and this "serve" subcommand seems to work:

```bash
civibuild create localhost --type standalone-dev --url 'http://localhost:8000'
civibuild serve localhost
```

So in theory, the following process should work with this PR:

```bash
git clone https://github.com/civicrm/civicrm-buildkit ~/buildkit
cd ~/buildkit
./bin/civi-download-tools
amp config:set --mysql_dsn='mysql://root:foobar@127.0.0.1:3306'

civibuild create localhost --type standalone-dev --url 'http://localhost:8000'
civibuild serve localhost
```

totten:
@artfulrobot Well, it registers the MySQL connection...

With regard to everything else:

* As written, that command leaves all other settings alone - e.g. new deploys inherit [default values](https://github.com/amp-cli/amp/blob/0.7.3/app/defaults/services.yml#L9-L57). These would probably _work_, but another command might be _better_.
* The current defaults do *stuff*. It's *stuff* that doesn't really matter.
    * `httpd_type=apache` and `httpd_restart_command=` means: "Write out some vhost files in `~/.amp`, but they're not loaded, so they're irrelevant."
    * `perm_type=worldWritable` means: "Use `chmod 1777`, which runs everywhere. It's not [ideal](https://github.com/amp-cli/amp/blob/master/doc/perm.md#configuring-amp), but it doesn't impede your ability to do local dev."
* The better default would be like:
    ```bash
    amp config:set --mysql_dsn='mysql://root:foobar@127.0.0.1:3306' \
      --httpd_type=none --perm_type=none --host_type=none
    ```

totten:
> I've tried to r-run this but have not been successful so far.

Part of the issue may be that my subsequent update changed the scope. They're 2 qualitatively different scopes:

1. There's the `app/config/` stuff (for `standalone-clean` and `standalone-dev`) -- this should work in any place that already supports `civibuild`.
2. There's the `php -S` (`civibuild serve`) stuff -- i.e. new support for a different HTTP. This starts from different premises about the environment.

IMHO, the main blocker would be if part 1 caused problems for your regular environment.

> I can't run it in my/@michaelmcandrew 's [docker](https://github.com/michaelmcandrew/civicrm-buildkit-docker/) environment (although I've seen I'm using an old variant) because using amp tells Apache to start listening on that port, so serve fails because the port is in use.

Ah, right 🤦 . I can see how that might've been an interesting way to test part 2. That is a good reason for it to not work. (ie `civicrm-buildkit-docker` already has a better HTTPD... so there's no need for `php -S`...)

My main question would be on part 1 -- do `standalone-clean` or `standalone-dev` work in your regular environment?

> I also wonder whether it will fail because I'm sure I saw code in amp that assumes apache is present (it isn't here).

1. In one critical way, the overall system relies on Apache --`.htaccess` files. These are included in all major PHP apps+frameworks, and they're supported by Apache. Systemically speaking, this makes it much easier to work with Apache.
2. `amp` has a mode to manage/integrate Apache, so there will be references to it in `amp.git`.
3. `amp` doesn't require this mode. There was experimental support for nginx. There's an option `httpd_type=none`. On my local day-to-day systems, I run Apache separated from amp. (The file-path for `VirtualDocumentRoot` matches-up, but amp doesn't manage any services.)

> `ERROR: Missing main script (/download.sh)`

This is the same error that I get with `civibuild create foobar --type foobarbaz` -- i.e. it means that it failed to find the configuration folder (eg `app/config/foobarbaz` or `app/config/standalone-dev`).

If you're juggling multiple environments, then it could simply be that the the PR wasn't checked out on that copy? Or it could be that some env-vars/file-paths got criss-crossed? (e.g. The call to `civibuild` was meant for enviroment-A but spliced stuff from environment-B.)

---

In any event, I think main question should be `standalone-{clean,dev}` (part 1). The `php -S`/`serve` stuff (part 2) can be more of a hidden/pending thing for now.

PR #793, opened 2023-07-25
Remove devel and volunteer from D7 builds

mlutfy:
The Drupal 7 sandbox is failing because of devel:

```
There are no stable releases for project devel.                        [warning]
Choose one of the available releases for devel:
 [0]  :  Cancel                                      
 [1]  :  7.x-1.x-dev  -  2020-Feb-20  -  Development 
 [2]  :  7.x-1.7      -  2019-Feb-22  -  Recommended 
 [3]  :  7.x-1.1      -  2011-Jul-20  -  Security
Cancelled.                                                              [cancel]
```

I'm sure it's an easy fix, but there's no need for it anyway.

I also removed CiviVolunteer while I was there.

### DIFF
--- a/app/config/backdrop-demo/download.sh
+++ b/app/config/backdrop-demo/download.sh
@@ -26,8 +26,6 @@ pushd "$WEB_ROOT/web/modules" >> /dev/null
   git clone ${CACHE_DIR}/civicrm/civicrm-core.git      -b "$CIVI_VERSION"     civicrm
   git clone ${CACHE_DIR}/civicrm/civicrm-backdrop.git  -b "1.x-$CIVI_VERSION" civicrm/backdrop
   git clone ${CACHE_DIR}/civicrm/civicrm-packages.git  -b "$CIVI_VERSION"     civicrm/packages
-  git clone "${CACHE_DIR}/civicrm/civivolunteer.git"                       -b "$VOL_VERSION"       civicrm/tools/extensions/civivolunteer
-  git clone "${CACHE_DIR}/ginkgostreet/org.civicrm.angularprofiles.git"    -b "$NG_PRFL_VERSION"   civicrm/tools/extensions/org.civicrm.angularprofiles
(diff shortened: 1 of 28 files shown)


totten:
@mlutfy Yeah, I think it's fine policy to remove `devel`. Should we remove it from the D8/D9/D10 builds?

I'm personally on the fence about removing `volunteer`. But if it's removed, then the current patch doesn't quite work. I ran this command:

```bash
civibuild create tmp --type drupal-demo
```

And the build fails after this step:

```
+++ [[ civirules civisualize cividiscount org.civicrm.search_kit org.civicrm.search org.civicrm.contactlayout org.civicrm.angularprofiles org.civicrm.volunteer civicrm_admin_ui message_admin org.civicrm.afform org.civicrm.afform_admin =~ volunteer ]]
+++ drush scr /Users/totten/bknix/src/drush/perm.php
Could not find the permission: register to volunteer                 [error]
Could not find the permission: register to volunteer                 [error]
Could not find the permission: register to volunteer                 [error]
```

For removing it, I would suggest these other changes:

1. Update `src/civibuild.defaults.sh:CIVI_DEMO_EXTS=...`. (*This disables it on all build-types.*)
2. To save space, stop downloading on all demo build-types. (Grep `tools/extensions/civivolunteer` to find the other download steps.)
3. To simplify code, drop the bits [like this](https://github.com/civicrm/civicrm-buildkit/blob/c72c28aede9e026f5f1211572444bfb7a3c64df8/app/config/drupal-demo/install.sh#L140-L158) for all demo build-types. (*These bits could hypothetically be retained in case someone wants to locally customize `CIVI_DEMO_EXTS`.*)

mlutfy:
@totten Thanks - I went ahead and did more cleanup for Volunteer, and tested locally for dmaster and `wp-demo`.

I did some cleanup for some non-supported builds (d8/civihr) to make grep happy, but didn't test.

and a minor update to the git cache for `civisualize` which is now on Gitlab.

totten:
Cool, changes look to do as advertised. I spot-checked a few builds and they ran OK. Added a couple more related cleanups. Merging.

PR #794, opened 2023-07-25
src/civibuild.aliases.sh: add smaster for c-i demo builds

mlutfy:
Adds an `smaster` alias so that we can run:

```
civibuild create smaster --url http://example.org
```

which is what Jenkins wants when rebuilding the demo sites (https://test.civicrm.org/view/Sites/job/demo.civicrm.org/).

I added the DNS record, and updated Jenkins, but only tested locally.

side-note: I removed "dcase" in Jenkins, because it was broken and not used.


totten:
Woot, `demo.civicrm.org`

mlutfy:
almost works, but I hit some issues (file permissions, https). Detailed here: https://lab.civicrm.org/infra/ops/-/issues/998

PR #800, opened 2023-07-27
Pass ExtensionsDir setting to core:install  + WMF install fix

eileenmcnaughton:


eileenmcnaughton:
Merging this as the WMF bit is in my -merge-at-will space and the other part is a patch @totten proposed which I tested

eileenmcnaughton:
(also so I can check if it works on our CI & if not I will revert the WMF bit)

PR #804, opened 2023-08-01
standalone: give demo user default 'staff' role

artfulrobot:
Requires https://github.com/civicrm/civicrm-core/pull/26979

After this and that ↑ 
you should be able to login as user `demo` and see Civi.


seamuslee001:
It has activity eh github

totten:
@artfulrobot woot

PR #808, opened 2023-08-18
Adapt new phars script for possibility of git-checked-out civix

eileenmcnaughton:
This is a pretty unusual config - but it's also a pretty small accomodation to make it work...

### DIFF
--- a/src/Civi/Buildkit/Phars.php
+++ b/src/Civi/Buildkit/Phars.php
@@ -100,11 +100,12 @@ class Phars {
       throw new \Exception(sprintf("Download for \"%s\" failed SHA-256 check (expect=%s, actual=%s)", $remoteUrl, $expectHash, $actualHash));
     }
 
-    if (file_exists($localFile)) {
+    // Civix could be a directory if checked out as a repo (unusual but possible configuration).
+    if (file_exists($localFile) && !is_dir($localFile)) {
       unlink($localFile);
     }
 
-    if (!rename($tempFile, $localFile)) {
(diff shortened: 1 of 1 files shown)


src/Civi/Buildkit/Phars.php
```diff
@@ -100,11 +100,12 @@ protected static function downloadFile(string $remoteUrl, string $expectHash, st
       throw new \Exception(sprintf("Download for \"%s\" failed SHA-256 check (expect=%s, actual=%s)", $remoteUrl, $expectHash, $actualHash));
     }
 
-    if (file_exists($localFile)) {
+    // Civix could be a directory if checked out as a repo (unusual but possible configuration).
```
totten, line 103: So you're file-layout is like this?

```
BUILDKIT/bin
BUILDKIT/bin/amp
BUILDKIT/bin/civix/bin/civix
```

That seems like a weird place to put the folder.

But, OK, if you're skip it... then shouldn't it skip the entire download? Otherwise, you're just gonna keep re-running 2mb HTTP requests and never using the content.

totten: I'll do an update for ^^

eileenmcnaughton: Well I don't call download-tools very often.....

totten: @eileenmcnaughton OK, try this:  https://github.com/civicrm/civicrm-buildkit/pull/809

PR #810, opened 2023-08-21
Stop WP from doing a Major upgrade on install

kcristiano:
If we specify a version - keep that version.

Problem:

Install wp-demo and specify a version  `--cms-ver 6.2.2`  

Buildkit installs Version 6.2.2

Launch the site and login

WP updates to WP latest (currently 6.3)

Solution:

Set `WP_AUTO_UPDATE_CORE` to only auto-update  minor versions.


totten:
That makes sense! Thanks @kcristiano

PR #811, opened 2023-08-23
Use PHPunit9 on CiviCRM versons 5.56 and more recent

seamuslee001:
@totten @colemanw

### DIFF
--- a/bin/civi-test-run
+++ b/bin/civi-test-run
@@ -14,14 +14,16 @@ function getPhpunitVer() {
         return
         ;;
       phpunit9)
-        echo "$PHPUNIT --debug"
+        echo "$PHPUNIT --printer \Civi\Test\TAP"
         return
         ;;
     esac
   fi
 
   CIVIVER=$(getCiviVer)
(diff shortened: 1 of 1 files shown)


colemanw:
I feel like all that code could be simplified if we just drop PHPUnit5

totten:
Looks pretty sensible. I'm giving it a trial run of `phpunit-core-exts` to make sure they work on on phpunit9.

totten:
Test results looked fine on phpunit9, notwithstanding minor notices. (*Tested with `master` + D7 + `min`imum requirements.*)

> I feel like all that code could be simplified if we just drop PHPUnit5

Yeah, probably. For `civi-test-run`, I think it's OK to drop PHPUnit5. It's pretty rare to need to run <=5.22 (i.e. 43 month branch). And if you do back that far, then you probably need to switch around the whole work environment (*switch to an older flavor of buildkit*).

I'm a little on fence about whether to actually remove the `phpunit5.phar`. It is fairly old. OTOH, it had an extremely long run as the recommended version for extensions.

PR #821, opened 2023-11-07
Enable unit testing using Smarty3

seamuslee001:
This aims to enable unit testing on PR test jobs but leaving smarty2 enabled on matrix jobs however you can enable it by doing export SMARTY3_ENABLE=true

ping @totten @eileenmcnaughton

### DIFF
--- a/src/pogo/civi-test-pr.php
+++ b/src/pogo/civi-test-pr.php
@@ -64,6 +64,7 @@ $c['app']->command("main [-N|--dry-run] [-S|--step] [--type=] [--patch=] [--keep
   $taskr->passthru('civibuild install {{0|s}}', [$c['buildName']]);
 
   $io->section("\nRun tests");
+  $taskr->passthru('export SMARTY3_ENABLE=true');
   $taskr->passthru('TIME_FUNC={{3|s}} civi-test-run -b {{0|s}} -j {{1|s}} {{2}}', [
     $c['buildName'],
     $c['junitDir'],
--- a/app/civicrm.settings.d/400-smarty3.php
+++ b/app/civicrm.settings.d/400-smarty3.php
@@ -1,18 +1,26 @@
 <?php
 
-switch($GLOBALS['_CV']['CIVI_UF']) {
-  case 'Drupal':
-  case 'Backdrop':
-  case 'WordPress':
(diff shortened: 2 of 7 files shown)


eileenmcnaughton:
@seamuslee001 do we know if it passes a full run?

seamuslee001:
@eileenmcnaughton no idea as this hasn't been done on any of the test boxes yet

eileenmcnaughton:
@seamuslee001 well we can always revert - I think merging it takes you forward on getting testing going

src/pogo/civi-test-pr.php
```diff
@@ -64,6 +64,7 @@
   $taskr->passthru('civibuild install {{0|s}}', [$c['buildName']]);
 
   $io->section("\nRun tests");
+  $taskr->passthru('export SMARTY3_ENABLE=true');
   $taskr->passthru('TIME_FUNC={{3|s}} civi-test-run -b {{0|s}} -j {{1|s}} {{2}}', [
```
totten, line 68: This change shouldn't break anything, but I suspect it doesn't do what you want:

* On L67, `passthru()` would open a new `bash` subprocess (PID `100`), execute the statement, and then close that process. 
* On L68, `passthru()` would open a new `bash`subprocess (PID `200`), execute the statement, and then close that process.
* But those two processes (PID `100` and PID `200`) have separate environments. PID 200 doesn't inherit the environment from PID 100.

I think what you're trying to write is either:

```php
// Pass the variable directly to the subprocess
$taskr->passthru('SMARTY3_ENABLE=true TIME_FUNC={{3|s}} civi-test-run ...');
```
```php
// Or set the variableon the parent process... before spawning children.
putenv('SMARTY3_ENABLE=true');
$taskr->passthru('TIME_FUNC={{3|s}} civi-test-run ...');
```
```php
// Well, ^^ is the obvious thing. Real-world PHP actually requires:
$_SERVER['SMARTY3_ENABLE'] = $_ENV['SMARTY3_ENABLE'] = 'true'; 
putenv('SMARTY3_ENABLE=' . $_SERVER['SMARTY3_ENABLE']);
$taskr->passthru('TIME_FUNC={{3|s}} civi-test-run ...');
```

PR #822, opened 2023-11-07
wp-clean - Do not use latest cutting edge theme

kcristiano:
Don't enable default FSE Theme


totten:
Thanks @kcristiano. I did a sample build locally and it appeared to work. 👍

PR #833, opened 2023-12-12
standalone-dev build: Correct upload path (upload not uploads)

artfulrobot:
This has been bugging me for ages!

Install script created web/uploads instead of web/upload (resulting in an uploads/ dir *and* an upload/ dir)

### DIFF
--- a/app/config/standalone-dev/download.sh
+++ b/app/config/standalone-dev/download.sh
@@ -7,10 +7,10 @@
 [ -z "$CMS_VERSION" ] && CMS_VERSION=master
 ## Hmm, not really used...
 
-mkdir -p "$WEB_ROOT" "$WEB_ROOT/web" "$WEB_ROOT/web/uploads" "$WEB_ROOT/data"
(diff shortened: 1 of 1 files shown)


app/config/standalone-dev/download.sh
```diff
@@ -7,7 +7,7 @@
 [ -z "$CMS_VERSION" ] && CMS_VERSION=master
 ## Hmm, not really used...
 
-mkdir -p "$WEB_ROOT" "$WEB_ROOT/web" "$WEB_ROOT/web/uploads" "$WEB_ROOT/data"
+mkdir -p "$WEB_ROOT" "$WEB_ROOT/web" "$WEB_ROOT/web/upload" "$WEB_ROOT/data"
 
 pushd "$WEB_ROOT"
   amp datadir "./data" "./web/uploads"
```
totten, line 13: This should also match, no?

artfulrobot: yup. done.

PR #834, opened 2023-12-13
Git Caches - Only clone/update repos that are actually being used

totten:
Before
---------

All builds load `app/config/caches.sh`. This script globally and unconditionally prefetches a dozen git repos. 

In each `app/config/*/download.sh`, you have statements that rely on prefetching, such as:

```bash
git clone "$CACHE_DIR/civicrm/civicrm-core" "$WEB_ROOT/sites/all/modules/civicrm"
```

After 
-------

All builds load `app/config/caches.sh`, but this script is now empty. There is no global/unconditional prefetch.

In each `app/config/*/download.sh`, you have statements like:

```bash
gitit_cache_setup_id civicrm/civicrm-core
git clone "$CACHE_DIR/civicrm/civicrm-core" "$WEB_ROOT/sites/all/modules/civicrm"
```

For common repos (like `civicrm-core`), this can be written more concisely:

```bash
git_cache_clone  civicrm/civicrm-core "$WEB_ROOT/sites/all/modules/civicrm"
```

Motivation
-----------

Periodically (a few times each day or each week), the PR test-suite raises a false-negative because of a network issue. For example, a request to `lab.civicrm.org` or `github.com` fails. Sometimes the failures are necessary (*you can't test a patch you can't fetch!*), but I've noticed several in recent months which weren't. Rather, the cache was overzealous in prefetching git-repos.

For example, the typical profile used by PR tests (`drupal-clean`) doesn't need any extra extensions -- yet the prefetch mechanism polls `cividiscount`, `civirules`, and others from `lab.civicrm.org`.  (*These exts may be used in other build-types.. but `drupal-clean` is what matters for PR testing, and it doesn't need them...*) Polling unnecessary packages (*polling more web-hosts*) increases the odds that a random network-failure will cause trouble.

QA
-----------

I ran `civibuild download ...` with just about every modified profile (*and provoked some kind of cache-update for each*). They seemed to work as well as before.

Buildkit Forks
----------------

If you have a fork of `civicrm-buildkit` that adds or modifies `app/config/*/download.sh`, and if you directly reference git data from the `$CACHE_DIR`, then you may need to update your configurations.

The quick-and-dirty update is to re-enable the global/unconditional prefetch. Edit `app/config/caches.sh` and give a list of all the common/expected repos:

```bash
git_cache_setup_id civicrm/civicrm-{core,packages,backdrop,drupal,drupal-8,joomla,wordpress}
```

The slightly better option is to update each `app/config/*/download.sh` file which you manage. At the start of each file, setup the caches that you actually need, e.g.

```bash
git_cache_setup_id civicrm/civicrm-core civicrm/civicrm-packages
```


totten:
Updated description with more through "Before"/"After" and some QA notes. Added suggestions in case anyone has forked `civicrm-buildkit` with customizations in `app/config/*/download.sh`.

PR #839, opened 2024-01-15
add some selected systopia repos

jofranz:
This adds some selected repositories to buildkit in a sorted way


totten:
Nice. Thanks @jofranz

PR #844, opened 2024-02-29
bknix - Switch default storage policy to regular disk (not ramdisk) 

totten:
When using buildkit's nix configuration to run PHP/MySQL, it generates runtime data (such as MySQL databases, esp `ibdata`). This changes the policy.

* Before
    * __Default__: Store data on ramdisk
    * __Opt-in__: You can edit `loco.yml` to disable the ramdisk. (To enable regular disk.)
* After
    * __Default__: Store data on regular disk
    * __Opt-in__: You can set an env-var (`RAMDISK_SIZE`) to enable ramdisk. (*You can do this by editing `loco.yml`... or calling `export`... or editing `.profile`... ad nauseum.*)

### DIFF
--- a/.loco/loco.yml
+++ b/.loco/loco.yml
@@ -1,6 +1,12 @@
 format: 'loco-0.1'
 
 ## This is the default configuration used for local development environments.
+##
+## You can modify the configuration with any of these techniques:
+##  - Edit this file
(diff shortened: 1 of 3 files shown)


totten:
JFYI @colemanw, to continue using ramdisk after the update, you'll want to update local config. I'm using:

```bash
sudo mkdir /etc/bknix-ci
sudo nano /etc/bknix-ci/loco.overrides.yml
```

with content:

```yaml
environment:
  - RAMDISK_SIZE=650
```

PR #845, opened 2024-03-08
Change Edge PHP version to 8.3 and change max and dfl to be 8.2

seamuslee001:
@totten @eileenmcnaughton

### DIFF
--- a/nix/profiles/dfl/default.nix
+++ b/nix/profiles/dfl/default.nix
@@ -12,7 +12,7 @@ let
 
 in (import ../base/default.nix) ++ (import ../mgmt/default.nix) ++ [
 
-    dists.bkit.php81
+    dists.bkit.php82
     dists.default.nodejs-14_x
     dists.default.apacheHttpd
     dists.default.mailhog
--- a/nix/profiles/edge/default.nix
+++ b/nix/profiles/edge/default.nix
@@ -11,7 +11,7 @@ let
 
 in (import ../base/default.nix) ++ (import ../mgmt/default.nix) ++ [
 
-    dists.bkit.php82
+    dists.bkit.php83
     dists.default.nodejs-14_x
     dists.default.apacheHttpd
     dists.default.mailhog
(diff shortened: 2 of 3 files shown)


eileenmcnaughton:
thanks @seamuslee001

totten:
This is gonna be interesting to see what happens to the matrices...

totten:
Well... https://test.civicrm.org/job/CiviCRM-Core-Matrix/ ... the effect is not great... and not terrible...

As you'd probably expect, the effect on `5.72` and `master` is fairly small -- adding 2-3 failures within `phpunit-crm,max`. Seems OK since we have some time before release.

For `5.69`, it's throwing 68 failures on `phpunit-crm,max` and 4 on `phpunit-api3`. 70 is  enough to be a challenge to reliably reading the data. I guess we don't actually care about this matrix for 5.69 -- what matters is `Security-Core-Matrix` for `5.69`. (But that's gonna behave the same.) 

For `5.71`, it's throwing 6. Probably small enough that we could potentially forward port fixes.

colemanw:
This is great. I need to use php 8.2 locally (because developing Civix, `box` refuses to run on anything less).
However, after pulling this down, running `civi-download-tools` and restarting my computer a few times, I still am stuck with php 8.1.
I see this:
```bash
> use-bknix dfl -s
> php -v
PHP 8.1.19 (cli)
```
and it seems to be coming from a local profile:
```bash
> which php
/nix/var/nix/profiles/per-user/colemanwatts/bknix-dfl/bin/php
```
I don't remember ever creating a local profile and telling it to lock me on an old version of php, so I just tried deleting that directory:
```bash
> rm -R /nix/var/nix/profiles/per-user/colemanwatts/bknix-dfl
> use-bknix dfl -s
The specified profile does not correspond to an actual profile
```
Oops. Looks like I broke it.

colemanw:
Ok thanks for your help @totten. For future reference the fix required running these 2 commands
```sh
> ./nix/bin/install-developer.sh
> loco clean
```

PR #852, opened 2024-04-01
run-bknix-job: Work better in a world of many PHP/MySQL permutations

totten:
Overview
-------------

This updates the way that the test-runner loads PHP/MySQL. The overarching goal is to allow ephemeral testing with a more fluid mix of environments (PHP/MySQL/etc) - by relaxing a requirement to use a pre-installed permutation.

It's a dependency in the on-going update to the test-matrices, wherein `min`/`max` are replaced by a floating mix of environments like `php74m57`, `php83m80`, etc.

Technical Details
----------------------

A couple ways to look at the change:

* Internally, the job-script is written a little differently:
    * __Before__: The job-script said either `use-bknix` or `use_bknix_tmp`. Under the hood, this implied a call to `eval $(use-bknix $BKPROF)` -- which implied that the binaries were pinned/pre-installed profiles.
    * __Now__: The job-script says `assert_bknix_durable` or `assert_bknix_temporary`. But it doesn't have causal responsibility for finding/starting/stopping anything.
* Externally, the job-runner is invoked a little differently:
    * __Before__: The only way to run a job-script was `run-bknix-job max MyJob`. It used some hidden rules to decide the level isolation to apply to jobs.
    * __Now__: The responsibility is split, with different levels of isolation addressed as:
        * `run-job` simply runs the job-script with your current PHP/MySQL/etc. It has no opinion about the isolation mechanics.
        * `run-bknix-job` uses nix to setup PHP/MySQL/etc. It has a few alternative modes:
            * `run-bknix-job --isolate max` (Run job in a temporary/isolated environment. Use Linux namespaces for isolation.)
            * `run-bknix-job --autostart max` (Run job as the current/local user. Get binaries from nix and start the services.)
            * `run-bknix-job --active max` (Run job in a pre-installed/already-active profile.)

### DIFF
--- a/bin/run-job
+++ b/bin/run-job
@@ -1,0 +1,151 @@
+#!/usr/bin/env bash
+{ # start: run-job
+
+RUN_BKNIX_CLEANUP_FILES=()
+RUN_BKNIX_CLEANUP_FUNCS=()
+trap run_job_cleanup EXIT
+
+function run_job_help() {
+  echo "usage: run-job [--mock] [JOB_NAME]"
+  echo
+  echo "example: run-job --mock CiviRM-Core-Matrix"
+  echo
+  echo "tip: If <JOB_NAME> is omitted, and if you execute within a real Jenkins"
+  echo "environment, then it will use the active \$JOB_NAME."
+  echo ""
+  echo "tip: If <JOB_NAME> is an HTTPS URL, then it will fetch and run the"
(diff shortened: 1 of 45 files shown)


totten:
Spot-checked by replaying a few jobs on real worker nodes. They seemed OK. Also seemed OK when running jobs with `BKRPOF=php83m80`.

PR #854, opened 2024-04-18
app/config - WordPress profiles should use new installer

totten:
Overview
----------

Update the build scripts for `wp-clean` and `wp-demo`.

Before
-------

It sets up the Civi database with `cat sql/civicrm.mysql sql/civicrm_data.mysql ... | mysql` and `cp civicrm.settings.php.template ...` (etal).

After
-------

It sets up the Civi database with `cv core:install` (aka `Civi\Setup`).

Comments
-----------

(1) During 5.72-alpha/5.72-rc, we had a critical regression get into the Civi-WP installation process. It was only caught during the final run-down. CI should be using `cv core:install` (*which is more representative and which would've caught the problem sooner*).

(2) I did a local run, e.g.

```bash

civibuild create delme --type wp-demo
```

And it seemed to have trouble. (*This is why I split apart the similar changes for WP and BD (#853).*) It's odd - I'm sure it worked at one point, and it's 90% the same as the web-based installer. 

Needs some work to track-down. This was sitting around on my system uncomitted... So for now I'm posting the draft PR to flag that this needs work.

### DIFF
--- a/app/config/wp-demo/install.sh
+++ b/app/config/wp-demo/install.sh
@@ -36,7 +36,7 @@ fi
 CIVI_TEMPLATEC="${CIVI_FILES}/templates_c"
 CIVI_UF="WordPress"
 
-civicrm_install
+civicrm_install_transitional
 
 ###############################################################################
 ## Extra configuration
--- a/app/config/wp-case/install.sh
+++ b/app/config/wp-case/install.sh
@@ -23,7 +23,7 @@ CIVI_FILES="${WEB_ROOT}/wp-content/uploads/civicrm"
 CIVI_TEMPLATEC="${CIVI_FILES}/templates_c"
 CIVI_UF="WordPress"
 
-civicrm_install
+civicrm_install_transitional
 
 ###############################################################################
 ## Extra configuration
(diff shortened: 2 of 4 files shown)


app/config/wp-demo/install.sh
```diff
@@ -36,7 +36,7 @@ fi
 CIVI_TEMPLATEC="${CIVI_FILES}/templates_c"
 CIVI_UF="WordPress"
 
-civicrm_install
+civicrm_install_transitional
```
kcristiano, line 39: The issue I see is that we call this function and then `civicrm_install_cv` before we activate the plugin so we get fatal errors.  

An ugly fix is to add 
```bash
pushd "$CMS_ROOT" >> /dev/null
wp plugin activate civicrm
wp eval '$c=[civi_wp()->admin, "add_wpload_setting"]; if (is_callable($c)) $c();'
```

before we call `civicrm_install_transitional`

totten: I just wanted to say a belated thanks @kcristiano. I was testing a related problem a week ago, and this clue helped a lot. I've been meaning to dig into understand problem and how to bake-in the fix. (Notwithstanding the github bot, the issue is still in the hopper...)

totten: OK, I think if we merge https://github.com/civicrm/civicrm-core/pull/30946, then `civirm_install_transitional` and `civicrm_install_cv` will. work better.

totten:
Re-tested. Works with current `master` (*5.78.alpha*) and `5.77` (*beta*).

PR #858, opened 2024-05-08
loco - Add support for launching Chrome services

totten:
This branch speaks to two scenarios of running buildkit-nix:

* For desktop usage (*default configuration, `.loco/loco.yml`*), you may want to manage Chrome yourself. (Because it's an interactive thing... and because you may already be using Chrome as your browser...) But this gives examples for how to auto-launch debuggable or headless instances. 
* For continuous-integration (`.loco/worker-n.yml`), always start+stop chrome as part of the test-run.

The main counter-point to this patch... `karma` already uses an autolaunch behavior for Chrome (*assign random port; start as part of the test-suite; end with the test-suite*). It doesn't expect any other process-manager (like `loco`).

* From DX POV, karma's approach a bit friendlier. It gives you something that works "out of the box" in more environments. (*You don't need to manually launch, and the autolaunch works in other environments*)
* From efficiency POV, it's better if phpunit+karma use the same pattern. Either *both* rely on `loco` to start Chrome; or *both*  their own short-lived Chrome instances. Mixing the approaches means (*as in the current PR*) means that we may get two copies of Chrome running at the same time. (*Loco will start its version unconditionally. Then karma will conditionally add another instance.*)

### DIFF
--- a/.loco/loco.yml
+++ b/.loco/loco.yml
@@ -49,7 +49,8 @@ default_environment:
  ## LOCALHOST: Bind services to a local IP address.
  - LOCALHOST=127.0.0.1
 
- - CHROME_BIN=$(which "chromium")
+ - CHROME_BIN=$(find-chrome)
+ - CHROME_PORT=9222
 
 #### Mandatory environment settings
 environment:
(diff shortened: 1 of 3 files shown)


totten:
Tweaked to support `$CHROME_HOST` (in lieu of `$LOCALHOST`).

For the moment, we're gonna go ahead with the current approach (which launches Chrome on all test-runs).

PR #859, opened 2024-05-09
civibuild - Add support for option CIVICRM_L10N_SYMLINK

totten:
Overview
------------

Adds an option, `CIVICRM_L10N_SYMLINK`. If enabled, new builds will use a shared l10n folder (based on symlinks).

Before
---------

* Most builds make a new copy of the `l10n` folder, which is large.

After
------

* By default, most builds still make a copy of `l10n/`.
* If you opt into `CIVICRM_L10N_SYMLINK`, then `l10n/` will be a symlink.

Comments
-------------

* The build-types `drupal{9,10}-dev` were added recently but accidentally omitted l10n data. This adds support.
* Any builds using bkinx (`nix-shell`/`loco`) will have the default switched to `CIVICRM_L10N_SYMLINK=1`.

### DIFF
--- a/.loco/loco.yml
+++ b/.loco/loco.yml
@@ -62,6 +62,7 @@ environment:
  - NODE_PATH=$BKIT/node_modules:$NODE_PATH
  - PATH=$BKIT/bin:$BKIT/node_modules/.bin:$LOCO_PRJ/.loco/bin:$PATH
  - CIVICRM_COMPOSER_ASSET=symdir
+ - CIVICRM_L10N_SYMLINK=1
 
 volume:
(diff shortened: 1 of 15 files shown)


totten:
I did some `r-run` with just about every build-type that was modified (alternately with and without `CIVICRM_L10N_SYMLINK`). In all cases, it toggled between concrete folders and symlinks. With either flag, translations worked in both web UI and install-data.

totten:
> The build-types drupal{9,10}-dev were added recently but accidentally omitted l10n data. This adds support.

Fun fact: This should fix some failures of `testLocalizedData()` in the periodic test-matrices.

PR #860, opened 2024-06-05
Update Standalone directory structure assumptions and rename versions

ufundo:
- updates directory structure assumptions to match https://github.com/civicrm/civicrm-core/pull/29960
- renames the two standalone variants: 
    - from `standalone-dev` to `standalone` as this is the tarball structure which we are saying will be the default (it's also easier for devs :))
    - from `standalone-clean` to `standalone-composer` as this is the composerfied variant which we don't expect people to use as much


totten:
For new `standalone-clean` (*simple layout*), I was getting failures like this:

```bash
civibuild create s3cln --type standalone-clean --patch https://github.com/civicrm/civicrm-core/pull/29960
```
```
...
Compiling additional files (For full details, use verbose "-v" mode.)
Compile: Generate CCL wrapper functions
Compile: Greenwich CSS (dist/bootstrap3.css)
~/bknix ~/bknix/build
[civibuild_apply_user_extras] Already applied
[[Install s3cln/default (type 'standalone-clean' in '/Users/totten/bknix/build/s3cln')]]
[[Execute /Users/totten/bknix/app/config/standalone-clean/install.sh]]
[[Setup MySQL and HTTP for CMS]]
[[Setup MySQL for Civi]]
[[Setup MySQL for Test]]
/Users/totten/bknix/app/config/standalone-clean/install.sh: line 22: pushd: /Users/totten/bknix/build/s3cln/core: No such file or directory
```

Committed an update which fixes that. The resulting site seems to work for (a) general spot-checking web UI, (b) running `cv`, and (c) running `phpunit`. I tried doing some file-uploads into the relocated folders, and...

* For uploading to custom-fields of type "File", it worked well. The upload landed in the private tree.
* For uploading via ckeditor/kcfinder, it failed. However, I'm not really sure if this worked before. I got a little bit further by replacing `web/core/civicrm.config.php` (aka `GENCODE_CONFIG_TEMPLATE`) with:

    ```php
    <?php
    chdir(dirname(__DIR__));
    require_once dirname(__DIR__) . '/civicrm.standalone.php';
    ```

    But kcfinder still has trouble. I think it's OK to leave that for a follow-up PR.

---

Next-up: testing `standalone-composer`

totten:
For this:

```bash
civibuild create s3com --type standalone-composer --patch https://github.com/civicrm/civicrm-core/pull/29960
```

I see:

```

        "adminPass": "admin",
        "adminEmail": "admin@example.com",
        "adminPassWasSpecified": false
    },
    "moFiles": {
        "civicrm.mo": "https://download.civicrm.org/civicrm-l10n-core/mo/[locale]/civicrm.mo"
    },
    "deployID": "YmsxRkmazIdhaw"
}
RUNNING: cvutil_inject_settings  civicrm.settings.d
missing variable: FILE [in cvutil_inject_settings]
```

The relevant part of `civibuild.lib.sh` is:

```bash
  cv core:install -vv -f --cms-base-url="$CMS_URL" --db="$CIVI_DB_DSN" -m "siteKey=$CIVI_SITE_KEY" "${installOpts[@]}"
  local settings=$( cv ev 'echo CIVICRM_SETTINGS_PATH;' )
  cvutil_inject_settings "$settings" "civicrm.settings.d"
```

The call to lookup `CIVICRM_SETTINGS_PATH` seems to fail. 

```bash
cd s3com/web
cv ev 'echo CIVICRM_SETTINGS_PATH;' ; echo exitcode=$?
```
```
exitcode=255
```

And every other `cv` command fails. So it seems to me like this arrangement isn't working (despite the recent cv#193).

ufundo:
Thanks for fixing the clean one :+1: 

For the composer one, there's an accompanying PR on the composer template repo: https://github.com/civicrm/civicrm-standalone/pull/11  - wonder if lack of that is why it's failing for you?

( Because the bootstrapping now specifies where to find the autoloader directly rather than searching a bunch of places, the composer arrangement requires a different version of `civicrm.standalone.php` . Suggested options here https://github.com/civicrm/civicrm-core/pull/29960#issuecomment-2109870251 and went with put the composer layout specific config in the composer template repo. )

It worked for me in non-buildkit context, but I wasn't able to orchestrate all the patches to test it with buildkit :grimacing:

ufundo:
So... the composer version needed updating with the CIVICRM_SETTINGS_PATH bit; and  https://github.com/civicrm/civicrm-standalone/pull/11 switches to the same default folder structure for the tarball for non-code folders. 

With tweaks here there, this is now working for me:

```
civibuild create soc --type standalone-composer --cms-ver ufundo/index-paths --patch https://github.com/civicrm/civicrm-core/pull/29960
```

Looks like these 3 outstanding PRs are dependent on each other now in a buildkit context. 

There's also switching the test build recipe name (if they are using `standalone-dev`? if using `standalone-clean` then can we keep tests using that (now tarball rather than composer structure) if that's our recommended installation route?) 

@totten

PR #863, opened 2024-06-25
civibuild - Allow running upgrade-test on Standalone

totten:
Overview
------------

Upgrade-tests ensure that you can... run upgrades... in different environments. `civibuild upgrade-test`  (aka `civibuild ut`) can be used like:

```
civibuild create build-0 --type standalone-clean --civi-ver master 
civibuild ut build-0 /home/totten/bknix/vendor/civicrm/upgrade-test/databases/5.47.2-tz.mysql.bz2
```

This is a step toward upgrade-tests on Standalone.

Before
--------

Upgrade tests don't run on standalone. For example [this periodic test](https://test.civicrm.org/duderino/batch/CiviCRM-Dev-Periodic/20868034) fired [this task](https://test.civicrm.org/job/CiviCRM-Test-QLow/33228/console) which outputs:

```
[[Load saved options from /home/homer/buildkit/build/build-0.sh]]
Skipped. civicrm-upgrade-test currently requires Drupal 7 or Backdrop and Drush.
```

After
------

You aren't specifically blocked. But it doesn't actually work, either.

```
[[Load saved options from /home/totten/bknix/build/build-0.sh]]
Identify files: [ "/home/totten/bknix/vendor/civicrm/upgrade-test/databases/5.47.2-tz.mysql.bz2"]
/home/totten/bknix/vendor/civicrm/upgrade-test/databases/5.47.2-tz.mysql.bz2
Web Dir: [/home/totten/bknix/build/build-0/web]
Output Dir: [/home/totten/bknix/build/build-0/.civibuild/debug]
Database Name: [build0civi_7x1h1]

In Standalone.php line 519:
                                              
  Class "Civi\Standalone\Security" not found  
                                              

path [--out OUT] [--flat [FLAT]] [-T|--out=table] [-I|--out=list] [--columns COLUMNS] [-x|--ext EXT] [-c|--config CONFIG] [-d|--dynamic DYNAMIC] [-m|--mkdir] [--level LEVEL] [--hostname HOSTNAME] [-t|--test] [-U|--user USER]
```

Comments
---------------

So... it still doesn't actually work. But it gets a step further!

The next step is going to be trickier. I *think* the problem here is actually about the content of the database:

* The upgrade-tests are based on [a library of database snapshots](https://github.com/civicrm/civicrm-upgrade-test/tree/master/databases).
* For Drupal/WordPress, the process is like this:
    * Install CMS/UF. (This goes into one database, eg `uf_db`).
    * Load snapshot. (This goes into another dbase, eg `civicrm_db`.)
    * Run `cv upgrade:db`
* But for standalone:
    * We only have one database. It contains both UF content and CRM content.
    * Whenever you load the snapshot, it resets the database.
    * Thus, loading `5.47.2-tz.mysql.bz2` destroys UF content (eg `standaloneusers`). You get an unusable system.
* Maintaining a snapshot library is a bit of work. (Which is why we don't update it very often...) So it seems useful for "Standalone" to leverage the same snapshot library as Drupal/WordPress. I suppose that means we need to find some way to splice together DB snapshot along with standalone's distinctive content. (*Maybe support split DB? Or maybe have some kind of filtering based on table-names? With something special for `civicrm_extension?*)
* This isn't the original goal, but there is a similar problem -- migration. You start out with a D7/WP site, backup the database to a file, and then load the database for a new standalone deployment. You need some way to fill-in the standalone pieces.

### DIFF
--- a/src/command/upgrade-test.run.sh
+++ b/src/command/upgrade-test.run.sh
@@ -1,5 +1,5 @@
 ## civicrm-upgrade-test only works with drush right now
-if grep -qi drupal "$CMS_ROOT/index.php" || grep -q backdrop "$CMS_ROOT/index.php" || [ -f "$CMS_ROOT/wp-config.php" ] ; then
+if grep -qi drupal "$CMS_ROOT/index.php" || grep -q backdrop "$CMS_ROOT/index.php" || [ -f "$CMS_ROOT/wp-config.php" ] || [ -f "$CMS_ROOT/civicrm.standalone.php" ] ; then
   cvutil_makeparent "$UPGRADE_LOG_DIR"
   cvutil_mkdir "$UPGRADE_LOG_DIR"
   pushd "$PRJDIR/vendor/civicrm/upgrade-test/databases" > /dev/null
@@ -7,5 +7,5 @@ if grep -qi drupal "$CMS_ROOT/index.php" || grep -q backdrop "$CMS_ROOT/index.ph
     ../bin/civicrm-upgrade-test --db "$CIVI_DB_NAME" --db-args "$CIVI_DB_ARGS" --web "$CMS_ROOT" --out "$UPGRADE_LOG_DIR" --junit-xml "$UPGRADE_LOG_DIR/civicrm-upgrade-test.xml" "${ARGS[@]:2}"
   popd > /dev/null
 else
-  echo "Skipped. civicrm-upgrade-test currently requires Drupal 7 or Backdrop and Drush."
+  echo "Skipped. civicrm-upgrade-test currently requires Drupal, Backdrop, WordPress, or Standalone."
 fi


totten:
> (@artfulrobot) I'm not sure how the snapshots are created?

It was more like:

1. Create a site (`civibuild create dmaster`). This has substeps:
    * Install CMS (`cms_db`)
    * Install Civi (`civicrm_db`)
2. Optionally, tweak the dataset by hand to exercise funny cases (like timezone snafu's or multilingual)
3. Make SQL snapshot of Civi content (`civicrm_db`)

This gives a bare/minimalist Civi DB snapshot, which can be re-used like:

1. Create site (`civibuild create wpmaster`). This has substeps:
    * Install CMS (`cms_db`) (*maybe the same CMS -- or maybe different CMS*)
    * Install Civi (`civicrm_db`) (*or not; we'll blow it away momentarily; the main thing is to get a `civicrm.settings.php` file initialized*)
2. Restore SQL snapshot of Civi (`civicrm_db`) 
3. Run upgrade (`cv upgrade:db` or `drush cvupdb` or similar)

The process has a compromise in that it doesn't really exercise `civicrm_uf_match` during upgrade; but it has a major upside in that we can quickly load various configurations on any supported UF.

> If it's an upgrade test, then can't we snapshot the result of a standalone install at version 5.x and use those and put in a thing to stop it trying to install an already-installed CRM? Too much data to do this for every x?

I can read this a couple ways, so I'm not certain I follow. Are you saying:

* Build larger/additional library of SQL snapshots. (It would be like [this folder](https://github.com/civicrm/civicrm-upgrade-test/tree/master/databases) but with snapshots specifically for standalone.)
    * That's conceivable.
* Don't have a library of SQL snapshots. Have a different upgrade-test routine which does a live installation of (say) 5.57 -- and then upgrades to the newer version.
    * (*Problems with this: it's slower, and install-requirements change over time. So you provision old install-requirements and then reprovision new install-requirements.*)

But just to put it out there... if we had some command for standalone -- like "import-snapshot" or "install-from-snapshot", which fills in standalone-specific-bits -- then that same command would be useful for migrations and for upgrade-testing. (*Upgrade-testing would literally be the same as migration-testing...*) And potentially, it would slot-in nicely (*don't need to reorganize test-suite*).

> I often attack problems by writing tests that fail first, then fixing them

Yeah, I know what you mean. For this specific situation, it's not particularly applicable.

ufundo:
Thanks for raising this @totten 

My initial reaction is with @artfulrobot - I think standalone-specific snapshots could be the way to go.

That keeps the support surface a bit tighter for Standalone - we assume Standalone installs start at at least 5.76(?), at least for now. (_If you are migrating from an older version, upgrade to 5.76 on your existing UF, hit any issues you might hit, then migrate_)

In sinilar vein I dont like the idea of supporting split DB, that feels like inviting problems which Standalone doesnt need to deal with.

I can see what you're saying about the use for an "add standalone bits" tool. But I think the priority use case should be for migration (preserving users) than a way to abstract the UF-like bits of Standalone from the rest. 

I have a general worry that if we always think of Standalone as just another UF we wont be able to really levwrage the benefits of integration, and we'll have another case to solve for UF-integration type problems. Maybe that's a bit abstract? But I think it would be reasonable to have slightly narrower range of possible config expectations for Standalone.

ufundo:
@totten picking this up again now - will merge this to get the test running - then look at how to get it passing

PR #865, opened 2024-07-17
standalone-composer - Fix support for `--civi-ver` flag

totten:
Overview
--------

I was recently surprised to see new test failures on the 5.75-stable test-matrix regarding Standalone... even though there were no relevant changes. The problem? It wasn't actually running tests on 5.75. It was running them on master (regardless of requested version).

This updates `standalone-composer` to respect the requested version.

Steps to Reproduce
------------------

```
civibuild create standalone-composer --civi-ver 5.75`
```

Before
------

It ignores the requested version (5.75) and installs `master`.

After
-----

IT respects the requested version. (But in absence of a version, it preserves the default `master`.)

### DIFF
--- a/app/config/standalone-composer/download.sh
+++ b/app/config/standalone-composer/download.sh
@@ -4,7 +4,7 @@
 
 ###############################################################################
 
-echo $CMS_VERSION
+[ -z "$CIVI_VERSION" ] && CIVI_VERSION=master
 [ -z "$CMS_VERSION" ] && CMS_VERSION=master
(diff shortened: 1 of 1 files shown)


totten:
Note: This doesn't actually make `standalone-composer` work with  `5.75`. There's a BC-break regarding installation steps between `5.75` <=> `5.76`, and this aligns with 5.76+. Regardless, the patch still seems a good idea so that (going forward) we can mix `standalone-composer` with `5.76`, `5.77`, etc.

PR #867, opened 2024-07-24
Add MySQL 8.4.1

totten:
Changes
----------
Add the package `mysql84` (8.4.1) from nixpkgs v24.05. 

Declare additional profiles:

* php74m84
* php80m84
* php81m84
* php82m84
* php83m84

Comments
-----------

@seamuslee001 In 8.4, they're phasing-out `default_authentication_plugin = mysql_native_password`. This doc (https://dev.mysql.com/doc/refman/8.4/en/mysql-nutshell.html#mysql-nutshell-deprecations) indicates that you can continue with another option (`mysql_native_password = ON`), so I put that in to get going.

But that seems like it's just overriding the deprecation. It's not clear what you're actually *supposed* to do instead...


totten:
replaced by #868

PR #868, opened 2024-08-01
(dev/core#5361) Add PHP 8.4.0.alpha3, MySQL 8.4.1, and MySQL 9.0.1

totten:
cc @seamuslee001 

Changes
----------

Add the packages `php84` (8.4.0.alpha3), `mysql84` (8.4.1), and `mysql90` (9.0.1). This uses nixpkgs v24.05 plus backports. Add new version of `loco` (0.8.0) fix warnings on php84.

Declare additional profiles involving permutations with these, such as

* `php74m84` (PHP 7.4 + MySQL 8.4)
* `php74m90` (PHP 7.4 + MySQL 9.0)
* ...
* `php83m84` (PHP 8.3 + MySQL 8.4)
* `php83m90` (PHP 8.3 + MySQL 9.0)
* ...
* `php84m80` (PHP 8.4 + MySQL 8.0)
* `php84m84` (PHP 8.4 + MySQL 8.4)
* `php84m90` (PHP 8.4 + MySQL 9.0)

Comments
-----------

* This PR is an alternative/replacement/expansion of #867.
* For PHP 8.4.0.alpha3, some PECL's are omitted due to build problems (eg `soap`, `imap`, `runkit7`). But those shouldn't block us from working on 8.4 compat.
* For MySQL 8.4 and MySQL 9.0, there's a major change to authentication. (See eg https://dev.mysql.com/doc/refman/8.4/en/mysql-nutshell.html#mysql-nutshell-deprecations)
    * For 8.4, the pre-existing mechanism (`mysql_native_password`) is deprecated but usable. I enabled it for 8.4, and it seems to work.
    * For 9.0, it drops the `mysql_native_password` mechanism. So it's using a newer mechanism (eg `caching_sha2_password`). If I understand correctly, this means that client-libraries (like PHP-PDO) need a corresponding update.
* Major issues coming next:
    * __php 8.4__: Various tools (like `drush`, `cv`, `amp`) are producing warnings. A lot of them are just warnings, but I am hitting errors with `drupal-empty`, `drupal-demo`, and `standalone-clean`.
    * __mysql 9.0__: Our current PHP-PDO drivers are having trouble connecting to MySQL 9.0:
    ```
    Fatal error: Uncaught PDOException: SQLSTATE[HY000] [2054] The server requested authentication method unknown to the client [caching_sha2_password] in phar:///home/totten/bknix/bin/amp/src/Amp/Database/Datasource.php:81
    ```
* But I am able to do a `civibuild reinstall dmaster` within `php83m84`. Yay.

These issues only affect the new versions -- so probably better to include them (and we can chip-away at issues while working toward compatibility).

This involves a bunch of new binaries. Suggested order of operations:

1. Approve changes
2. Send new binaries to binary-cache
3. Merge changes
4. Regenerate gcloud images


seamuslee001:
PHP 8 and up certainly should be fine with caching_sha2_password passwords I think

seamuslee001:
PDO reports it should be supported from 7.4.4 https://www.php.net/manual/en/ref.pdo-mysql.php

totten:
@seamuslee001  Yeah, the problem is using `IDENTIFIED WITH mysql_native_password BY ...` for provisioning. MySQL 9.0 seems to work fine if we go back to simpler `IDENTIFIED BY`...

It's not clear exactly when you *do* need `WITH mysql_native_password`. Maybe 8.0? In any case, working on test-suite update  for `amp` to get a more consistent view...

seamuslee001:
@totten I think it might have been MySQL 8.0 with earlier PHP7.4/7.3 versions or something?

totten:
@seamuslee001  Right, there are multiple cutoffs for different things. The prose in [the PHP doc](https://www.php.net/manual/en/ref.pdo-mysql.php) is about as clear as mud -- with key points at PHP 7.1.16, 7.2.4, and 7.4.4. :anguished: 

After revising and playing with the `amp` test suite, I think this might be a better way to explain it. (At least for me...) There are two questions: *client-server handshake compatibility* (will they talk to each other) and the *user-account compatibility* (will the specific user work).

| PHP Client Version | Handshake Compatibility | User Account Compatibility |
| -- |-- |-- |
| 7.1.0 | Server must default to `mysql_native_password` | Only users with `mysql_native_password` |
| 7.2.0 | Server must default to `mysql_native_password` | Only users with `mysql_native_password` |
| 7.1.16 | Server defaults are free to vary | Only users with `mysql_native_password` |
| 7.2.4 | Server defaults  are free to vary | Only users with `mysql_native_password` |
| 7.3.33 | Server defaults  are free to vary | Only users with `mysql_native_password` |
| 7.4.4 | Server defaults  are free to vary | Users with `caching_sha2_password` or `mysql_native_password` |

So it's a spectrum --  7.1.0 has the worst compatibility; the general range of 7.1.x-7.3.x is an improving mess; and finally 7.4.4 gets things straight.

totten:
Updated `amp` to v0.8.0. This was passing tests for mysql84/mysql90 (https://github.com/amp-cli/amp/pull/74) and php84 (https://github.com/amp-cli/amp/pull/76). It did require bumping up the minimum PHP version from 7.1 to 7.3, but I don't imagine anyone will have a problem with that.

__Note__: The specific combination of `php73` (*which requires `mysql_native_password`*) and `mysql90` (*which lacks `mysql_native_password`*) will probably never work. That's not really a CiviCRM issue, though. And I can't imagine anyone would eagerly jump up to `mysql90` while being locked ito `php73`. So *shrug*. We just won't have that combination in any of our matrices...

totten:
Note: Currently having a problem with publishing binaries. May have to continue looking at that a bit later...

totten:
Solved my problem publishing binaries.

It wasn't actually about php84 or mysql84 per se; rather, the PR had some reorganization of to promote DRYness (*`nix/profiles` now loops over the list of all php/mysql versions and makes profiles for them*); but this change meant that it was now visiting old packages (php56/php70) which had build errors. 

Dropped some old packages and did some more tidying. Now binaries are for Linux-x86 and Darwin-ARM.

PR #875, opened 2024-08-17
civihydra - Add support for standalone tarballs

totten:
Overview
----------------

`civihydra` is a variant of `civibuild`. It takes a list of release-files and generates new, empty sites for testing them.

This PR allows it to work with standalone.

Before
------------

It works with any mix of D7/WP/BD/J(?).

```
civihydra create /tmp/civicrm-X.Y.Z-drupal.tar.gz  /tmp/civicrm-X.Y.Z-wordpress.zip  /tmp/civicrm-X.Y.Z-backdrop.tar.gz
```

After
---------------

It also works with standalone

```
civihydra create /tmp/civicrm-X.Y.Z-standalone.tar.gz
```

Comments
---------------

This uses updated layout per https://github.com/civicrm/civicrm-core/pull/30904

### DIFF
--- a/bin/civihydra
+++ b/bin/civihydra
@@ -127,6 +127,14 @@ function hydra_create() {
         subsite_dl="web/tmp=$TARFILE"
         ;;
 
+      civicrm-*-standalone.tar.gz|civicrm-*-standalone-*.tar.gz)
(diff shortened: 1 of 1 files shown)


ufundo:
Works for me :+1:

PR #877, opened 2024-08-24
CI - Various improvements to process management

totten:
(Draft -- Should deploy to `test-4` and confirm that jobs are working there before we merge.)

### DIFF
--- a/bin/homerdo
+++ b/bin/homerdo
@@ -346,9 +346,10 @@
       local lock_pid=$(cat "$homer_lock")
       if [ -n "$lock_pid" ]; then
         if ps -p "$lock_pid" > /dev/null; then
-          echo >&2 "Image $img already in use by PID $lock_pid"
+          echo >&2 "Image $img already reserved. (Existing owner: $lock_pid. Rejected owner: $owner_pid.)"
           exit 2
         else
+          assert_unshare_inactive
           rm -f "$homer_lock"
           _task_unmount ## Consider moving out of critical section
         fi
@@ -388,6 +389,11 @@
     assert_valid_img; realize_img
 
     assert_ownership
+    assert_unshare_inactive
+    ## You could proceed with unmounting from root-context, but it's
+    ## misleading. It just means that root-context loses visibility to
+    ## the handles. The underlying resources persist...
+
     _task_unmount
   }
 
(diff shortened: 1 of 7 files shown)


totten:
This includes two main measures to mitigate the zombie pro`tblem. But first, some notes/context:

* Jenkins runs jobs by connecting over SSH, then starting a client, and then forking processes within that client.
* Here's an easy way to create a zombie: make an SSH connection to run a job and then `kill` your SSH client. 

    ```
    ssh dispatcher@localhost run-bknix-job --mock php74m80 shell
    ```
    ```
    ps uax|grep 'ssh dispatch' ...
    kill 1234
    ```

    The SSH server will shutdown the connection, but the processes within it will continue.
* The actual production situation is more complicated -- while it does rely on SSH, it doesn't make a new one for each job.  Rather, the Jenkins client has its own mechanism for creating+cleaning processes (*reportedly, [ProcessTreeKiller`](https://wiki.jenkins.io/display/JENKINS/ProcessTreeKiller)).
* There are different algorithms/data-structures for process cleanup. All that really matters here is that the `ProcessTreeKiller` algorithm doesn't seem to work with our scripts. (I haven't found a good way to trace this stuff, so an explanation beyond that would be speculative.)
* I spent a while trying to figure a way to add a little more PID-tracking to what we have (*"PID 123 is the parent; if it goes away, we should clean up child PID 456."*) However, I believe the substantive details -- which PIDs die at which moments -- differ between the basic SSH scenario (*which was easy to test*) and Jenkins scenario (*with ProcessTreeKiller*). So I wasn't with ability to test/confirm this procedure.

The PR implements two techniques which (*knock on wood*) should work the same in bare SSH or in Jenkins. The two techniques are complementary (*each fills a gap from the other*):

* `slotdo`: For each Jenkins `EXECUTOR_NUMBER`, this creates a transient systemd unit (`dispatcher_0.service`, `dispatcher_1.service`, etc). A job will normally end on its own (*and cleanup the service with it*). But if it doesn't (*if it's a zombie*), that's OK -- because the next job-run will force it to reset (`systemctl kill dispatcher_0.service`).
    *  The implementation here is called `slotdo` -- do work within a numbered slot (`slotdo run $EXECUTOR_NUMBER my-job.sh`). 
    * This depends on having another job come and do cleanup. But what if no other jobs are queued? You might have the zombie running for a long time.
* `homerdo --timeout 120m` (*2 hours*): The main thread of test will use GNU `timeout` to enforce a hard limit on its children. (AFAICS, all of these processes remain around in both bare-SSH and Jenkins-client arrangement.)
    * 2 hours should be ample time for normal jobs (*which are often 45-50m*).
    * If we're in the empty-queue situation (where `slotdo` won't  cleanup), then this timeout should eventually kick in. And if it spends 2 hours spinning a thread when the queue is empty... then OK.

So both mechanisms are imperfect, but both are better than what we've had, and they should compensate for each other's weaknesses.

totten:
Used `test-4` for the past fews as a trial run.

* Ran fine with normal loads for a couple days
* Tried more failure cases (tests which hang/timeout; tests which are manually killed), and they seemed to work fine.

PR #878, opened 2024-08-30
update wp-demo to include missing extensions

kcristiano:
Add changes to wp-demo to ignore missing ext dependencies and add back missing extensions.

@totten `git_cache_clone` did not work for me, do I need to update a different file to use that?

These are changes in my local environment that I think should be in the repo.

### DIFF
--- a/app/config/wp-demo/download.sh
+++ b/app/config/wp-demo/download.sh
@@ -27,6 +27,8 @@ pushd "$WEB_ROOT/web/wp-content/plugins" >> /dev/null
   git_cache_clone civicrm/civicrm-packages                         -b "$CIVI_VERSION" civicrm/civicrm/packages
   api4_download_conditional civicrm/civicrm                                              civicrm/civicrm/ext/api4
   git_cache_clone civicrm/civicrm-demo-wp                          -b master          civicrm-demo-wp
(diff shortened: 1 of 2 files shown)


totten:
> (@kcristiano) `git_cache_clone` did not work for me, do I need to update a different file to use that?

Yeah, it's a two-step process: setup the cache and then use the cache, eg

```bash
git_cache_setup https://example.com/my-repo "$CACHE_DIR/my/repo.git"
git_cache_clone my/repo -b TARGET_BRANCH ...
```

There are a couple of global rules ([a](https://github.com/civicrm/civicrm-buildkit/blob/master/src/civibuild.lib.sh#L1792-L1794), [b](https://github.com/civicrm/civicrm-buildkit/blob/master/app/config/caches.sh)) to setup common caches. These used to include CiviVolunteer (but not any more). So you'd have to add both statements back somewhere.

Personally, I don't have a strong feeling for or against enabling `civivolunteer`. But I should note that it was actually dropped from `drupal-demo` about a year ago (#793; cc @bgm), and it might nice if we weren't too bouncy on the question.

(*Possible compromise:. we could probably figure something involving `civibuild.conf` and /or `CIVI_DEMO_EXTS` to allow a more persistent opt-in -- so that the logic can be merged/committed without changing the global default..*)

kcristiano:
Thanks @totten 

I did not know it was removed from drupal-demo.  In that case I see no reason to add for WP.

PR #880, opened 2024-09-19
nix - Update php83 (8.3.9)

totten:


ufundo:
Doesn't fix my seg fault :thinking:

totten:
OK, well, I guess we should probably upgrade anyway.

PR #881, opened 2024-09-19
mysql - When running in nix env, disregard HOME/.my.cnf

totten:
When using `nix-shell` to start/stop MySQL, we create a new `my.cnf` (via `MYSQL_HOME=BUILDKIT/.loco/var/mysql/conf/`). 

This file is used in many new/clean environments. But if your system has an existing MySQL configuration (esp `$HOME/.my.cnf`), then this may prevent it from reading the generated one.

This revision ensures that (*within the nix-shell*) it will prefer the generated one. It does this putting a wrapper around `mysql`, `mysqladmin`, etc


totten:
Started seeing some errors in CI like [this](https://test.civicrm.org/job/CiviCRM-Style/2526/console):

```
       > /nix/store/28hqpbwpzvpff7ldbhxdhzcpdc34lgsa-stdenv-linux/setup: /nix/store/6y3gn6vj6v58x24r826csas1xhpkf00m-src/makeWrapperDir.sh: /usr/bin/env: bad interpreter: No such file or directory
```

(Partial revert: bcb25184495e392080e340aff344a1c41669bfa7)

PR #895, opened 2024-12-02
update standalone demo user creation

ufundo:
- create email record on the Contact (this should be propagated to user automatically)
- consistently set `admin` role (current standalone-clean sets admin, standalone-composer sets staff - I think arguments either way but taking standalone-clean as the canonical one for now
- simplify calls, use Api4 throughout

### DIFF
--- a/app/config/standalone-composer/demo-user.php
+++ b/app/config/standalone-composer/demo-user.php
@@ -15,27 +15,26 @@ foreach (['DEMO_EMAIL', 'DEMO_USER', 'DEMO_PASS'] as $var) {
 // Since users+contacts are stored in same DB, we can do transaction.
 CRM_Core_Transaction::create()->run(function () {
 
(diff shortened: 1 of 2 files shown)


app/config/standalone-composer/demo-user.php
```diff
@@ -15,27 +15,26 @@
...
-      'first_name' => 'Demo',
-      'last_name' => 'User',
-    ])
-    ->execute()->first()['id'];
-  $adminEmail = getenv('DEMO_EMAIL');
-  $params = [
-    'cms_name'   => getenv('DEMO_USER'),
-    'cms_pass'   => getenv('DEMO_PASS'),
-    'notify'     => FALSE,
-    $adminEmail  => $adminEmail,
```
ufundo, line 30: works, but weird... (also please dont pass `cms_pass` as your email...)

ufundo:
I dont think so. Would be good

PR #897, opened 2024-12-05
Install Riverlea in Demo sites and set the backend theme to be Minett…

seamuslee001:
…a (Greenwich)

ping @totten @vingle 

Note that this only kicks in from 5.80 on wards and this only applies to new builds using buildkit

cc @eileenmcnaughton @ufundo


totten:
Yup, adding that helper looks like a good approach. I spot-checked a couple local builds, and they worked. :+1: 

When RiverLea gets the WP-backend theme added, then we'll want to tweak the rules for `wp-demo` to use the other.  But we'll cross that bridge when we get to it.

vingle:
Parallel to this - is it possible to change the Drupal default theme to something other than Garland, as [that was removed from D8](https://www.drupal.org/node/911054) (and is quite narrow)? ie Claro or Seven

PR #900, opened 2024-12-10
php84: Update to 8.4.1 (etal)

totten:
Update php84 from 8.4.alpha to 8.4.1. Add updates for various PECLs that now work in 8.4.1.


totten:
New builds pushed up to binary-cache for linux-x64 and darwin-arm64.

PR #901, opened 2024-12-11
Ensure light mode is forced in backoffice on Drupal 7 builds

seamuslee001:
@ufundo @totten

### DIFF
--- a/src/civibuild.lib.sh
+++ b/src/civibuild.lib.sh
@@ -645,6 +645,9 @@ function civicrm_enable_riverlea_theme() {
   if civicrm_check_ver '>' 5.80.alpha1 ; then
     cv en --ignore-missing riverlea
     cv vset theme_backend=minetta
+    if [[ "$CIVI_UF" == "Drupal" ]]; then
+      cv vset riverlea_dark_mode_backend=light
+    fi
   fi
 }
 
--- a/src/civibuild.lib.sh
+++ b/src/civibuild.lib.sh
@@ -645,7 +645,7 @@ function civicrm_enable_riverlea_theme() {
   if civicrm_check_ver '>' 5.80.alpha1 ; then
     cv en --ignore-missing riverlea
     cv vset theme_backend=minetta
-    if [[ "$CIVI_UF" == "Drupal" ]]; then
+    if [[ "$CIVI_UF" == "Drupal" ] || [ "$CIVI_UF" == "WordPress" ]]; then
       cv vset riverlea_dark_mode_backend=light
     fi
   fi


kcristiano:
Thanks @seamuslee001  This looks like a needed patch.  I think it should be merged.

ufundo:
Thanks @seamuslee001 just r-running now

src/civibuild.lib.sh
```diff
@@ -645,6 +645,9 @@ function civicrm_enable_riverlea_theme() {
   if civicrm_check_ver '>' 5.80.alpha1 ; then
     cv en --ignore-missing riverlea
     cv vset theme_backend=minetta
+    if [[ "$CIVI_UF" == "Drupal" ] || [ "$CIVI_UF" == "WordPress" ]]; then
```
kcristiano, line 648: I needed to change to this:

```diff
diff --git a/src/civibuild.lib.sh b/src/civibuild.lib.sh
index 8ccd4fb..02a08e5 100644
--- a/src/civibuild.lib.sh
+++ b/src/civibuild.lib.sh
@@ -645,7 +645,9 @@ function civicrm_enable_riverlea_theme() {
   if civicrm_check_ver '>' 5.80.alpha1 ; then
     cv en --ignore-missing riverlea
     cv vset theme_backend=minetta
-    if [[ "$CIVI_UF" == "Drupal" ] || [ "$CIVI_UF" == "WordPress" ]]; then
+    if [ "$CIVI_UF" == "Drupal" ] ; then
+       cv vset riverlea_dark_mode_backend=light
+    elif [ "$CIVI_UF" == "WordPress" ] ;  then
       cv vset riverlea_dark_mode_backend=light
     fi
   fi
```

I was getting errors using the `||` syntax

kcristiano: with this setup - the back end setting is updated properly 

should we also set front end here?

totten:
Yeah, it regularly confuses me when I activate Riverlea locally.

The impact is a bit bigger than civibuild, though, right? Like... anyone who installs Civi on top of CMS can have the issue of (mis)matching light-dark mode? I guess I don't keep track of all themes in the CMS's, so I don't know what's statistically common... but I *feel* like it's more common for CMS themes to have a static color palette.

What about this policy (cc @vingle):

* When installing Riverlea on CiviCRM-Standalone, default to:
    `riverlea_dark_mode_{frontend,backend}=inherit`.
* When installing Riverlea on any other CMS, default to:
    `riverlea_dark_mode_{frontend,backend}=light`.

(*Or, if we can't do that, then we go to the next level -- update the web-installer to show a prompt.*)

vingle:
@totten I agree - it's definitely bigger than buildkit. Just yesterday had a client whose front-end Civi pages I've flipped to RiverLea commenting that they looked odd, and their screengrab showed it was their browser's dark-mode clashing with a front-end non-dark theme. So 100% yes to `riverlea_dark_mode_{frontend}=light`.

For backend, across the other CMS, afaict:
 - Backdrop - there's a 4yr old issue that got a PR to [add darkmode in Sept](https://github.com/backdrop/backdrop-issues/issues/4778), so looks like it comes soon.
 - Joomla 3 & 4 - no darkmode; 
 - Joomla 5 - darkmode
 - Drupal 7 Garland & Seven - no darkmode;
 - Drupal 10 Claro, Seven - no darkmode; 
 - Drupal 10 Gin & Bootstrap 5 - darkmode
 - WordPress - no darkmode
 
Joomla 5, D10 Gin/BS5, Backdrop at some point soon - def not the majority… So maybe it's safe to set them all to light by default and just be a bit more noisy about the darkmode and how to turn it on, and then revisit this if WordPress or Drupal default admin theme add it?

ufundo:
that list certaimly suggests `light` as a sensible default across the board.

then leaves open the possibility of setting inherit in the Standalone installer?

ufundo:
https://github.com/civicrm/civicrm-core/pull/31611 ?

vingle:
> then leaves open the possibility of setting inherit in the Standalone installer?

Yes - and then if/when Backdrop merges the darkmode PR, and when Joomla5 becomes 90% of Joomlas, could include them too? But never for frontend (unlike Standalone which has the same theme environment front & back).

ufundo:
I think this isn't needed now https://github.com/civicrm/civicrm-core/pull/31611 and https://github.com/civicrm/civicrm-core/pull/31633 are merged.

dmaster and smaster both looking good today

PR #904, opened 2025-01-14
Don't try to download libraries for demo drupal 7

demeritcowboy:
See https://chat.civicrm.org/civicrm/pl/x44sdeq8gi8ipc9yihoyjy743a. There's no stable release in whatever `drush dl` uses to find it.

The module isn't installed on the demo site anyway, so don't try to download it.


totten:
Thanks @demeritcowboy. Great point. I merged an expanded variant (to deal with more build-types and modules) in #907.

PR #908, opened 2025-01-21
Update Composer to latest LTS version 2.8.4

seamuslee001:


totten:
Spot-checked to make sure it works with PHP 7.4 (for outgoing 5.75-esr and incoming 5.81-esr). Both seemed good.

PR #909, opened 2025-01-27
Avoid unnecessary timeouts to gitlab causing drupal 10 public demo build fails

demeritcowboy:
The drupal 10 demo build has been failing the last 4 days timing out while connecting to gitlab. See https://chat.civicrm.org/civicrm/pl/nhwn4775y7gfpfwpgsepeaebih

In the line just above, it moves a local extension from one folder to another, so I assume the purpose of this refresh line is to pick up that change. It doesn't need to contact gitlab for that.

### DIFF
--- a/app/config/drupal10-demo/install.sh
+++ b/app/config/drupal10-demo/install.sh
@@ -69,7 +69,7 @@ pushd "${CMS_ROOT}/sites/${DRUPAL_SITE_DIR}" >> /dev/null
   if [ -d "$CIVI_CORE/tools/extensions/org.civicrm.contactlayout" ]; then
     mv $CIVI_CORE/tools/extensions/org.civicrm.contactlayout files/civicrm/ext
   fi
-  cv api extension.refresh
+  cv api extension.refresh local=1 remote=0
 
   ## Demo sites always disable email and often disable cron
   cv api StatusPreference.create ignore_severity=critical name=checkOutboundMail


stesi561:
Looks good. I've not r-run but have copy and pasted to test that command on a local install of CiviCRM just to be sure. 

I do wonder though if we need that preceding copy operation - I mean it's probably just a no-op. 

But my read of it is it's trying to find something that doesn't exist here?
https://github.com/civicrm/civicrm-core/tree/master/tools/extensions

Or does that get populated from elsewhere. I don't think that should be addressed here as it would complicate getting it merged.

app/config/drupal10-demo/install.sh
```diff
@@ -69,7 +69,7 @@ pushd "${CMS_ROOT}/sites/${DRUPAL_SITE_DIR}" >> /dev/null
   if [ -d "$CIVI_CORE/tools/extensions/org.civicrm.contactlayout" ]; then
     mv $CIVI_CORE/tools/extensions/org.civicrm.contactlayout files/civicrm/ext
   fi
-  cv api extension.refresh
+  cv api extension.refresh local=1
```
totten, line 72: Agree this idea seems fine.

When I first ran this, it still crashed on the remote download step. If you look at the 'api.default` for `Extension.refresh`, it specifies both `local=1` and `remote=1`.

The clearer version of this call would be like:

```suggestion
  cv api extension.refresh local=1 remote=0
```

That did fix for me.

(But when I went back to revalidate everything and understand the failing lookups... well, at that exact moment, the remote lookups seemed to work again...)

Anyway, I think it's a fine change since this script doesn't actually download extensions. So skipping the remote-refresh will trim some time+risk.

demeritcowboy: Oops. Updated.

In terms of understanding the fails, I believe it's this https://lab.civicrm.org/infra/extdir/-/issues/9, which has been exacerbated by the recent-ish change to always include "unapproved" extensions.

And @stesi561 yes I believe it does pull it in here: https://github.com/civicrm/civicrm-buildkit/blob/af5c0ec662461353dad5cb9ddd34b2f87444875e/app/config/drupal10-demo/download.sh#L32

stesi561: aha! I should have hit grep!

PR #911, opened 2025-02-06
Update buildkit for use with new site_email_address table

colemanw:
This complements https://github.com/civicrm/civicrm-core/pull/31909 by installing site email in the appropriate table depending on version.

### DIFF
--- a/app/config/backdrop-clean/install.sh
+++ b/app/config/backdrop-clean/install.sh
@@ -20,7 +20,7 @@ backdrop_install
 ## Setup CiviCRM (config files, database tables)
 
 CIVI_DOMAIN_NAME="Demonstrators Anonymous"
-CIVI_DOMAIN_EMAIL="\"Demonstrators Anonymous\" <info@example.org>"
+CIVI_DOMAIN_EMAIL="info@example.org"
 CIVI_CORE="${CMS_ROOT}/modules/civicrm"
 CIVI_SETTINGS="${CMS_ROOT}/civicrm.settings.php"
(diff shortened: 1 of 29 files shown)


app/config/backdrop-clean/install.sh
```diff
@@ -20,7 +20,7 @@ backdrop_install
 ## Setup CiviCRM (config files, database tables)
 
 CIVI_DOMAIN_NAME="Demonstrators Anonymous"
-CIVI_DOMAIN_EMAIL="\"Demonstrators Anonymous\" <info@example.org>"
+CIVI_DOMAIN_EMAIL="info@example.org"
```
totten, line 23: I'm a little torn. The above would be fine greenfield-wise, and it's written. OTOH, I try to keep env-vars pretty stable for anyone who has a fork with their own build-type. OTOOH, this isn't a contract that we're vocal about.

The underlying issue is probably that this kind of value pushes the limits of what's intuitive/safe for processing with shell commands. (Actually, `cv sql --eval` has a better notation `@ENV[XXX]`, but api3/api4 don't have an equivalent -- and api3/api4 seems seems like a nicer bridge -- eg when faced with `git bisect`, `cv api4 OptionValue.update` would endure better.)

I think I'll go ahead and merge this to keep 31909 moving. But I reserve the right to flipflop :wink:  (*if I find a nicer way to switch it to api4*).

colemanw:
@totten so you're saying there might be build types floating around with `CIVI_DOMAIN_EMAIL="\"Demonstrators Anonymous\" <info@example.org>"`?

What about just add a little regex here to extract the email part from the angle brackets?

PR #918, opened 2025-02-27
Update php8.5.27 to 8.5.41

eileenmcnaughton:
Not sure if this will get us past the deprecation notices but makes sense to do even if not

https://phar.phpunit.de/

> Deprecated: PHPUnitFrameworkTest::run(): Implicitly marking parameter  as nullable is deprecated, the explicit nullable type must be used instead in phar:///home/homer/buildkit/extern/phpunit8/phpunit8.phar/phpunit/Framework/Test.php on line 23

@totten @seamuslee001 not sure why we are using phpunit8 against php edge ... but this seems like a sane first step towards looking at the deprecations

https://test.civicrm.org/job/CiviCRM-Test-QLow/134517/console

### DIFF
--- a/phars.json
+++ b/phars.json
@@ -75,8 +75,8 @@
    },
 
    "phpunit8":  {
-     "url": "https://phar.phpunit.de/phpunit-8.5.27.phar",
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
thanks @seamuslee001 it was confusing - but this is OK to merge anyway to reduce noise?

PR #919, opened 2025-03-05
skip chromium for nix on all Apples, not just M1

ufundo:
It seems like Chromium builds aren't available for Intel Macs either, so I think the opt out here should be broaden.

As is, it caused quite a lot of headaches trying to setup on a fresh install of Nix on one of the last Intel MacBook Pros this morning. Skipping chromium it worked nicely.

There seem to be various closed PRs relating to adding/reinstating Chromium package for Darwin. The last mentions a `flake` that could maybe be used, though I don't know how that would work! https://github.com/NixOS/nixpkgs/pull/324701 ?

### DIFF
--- a/nix/profiles/phpXXmXX/default.nix
+++ b/nix/profiles/phpXXmXX/default.nix
@@ -9,7 +9,7 @@
 let
     dists = import ../../pins;
     stdenv = dists.default.stdenv;
-    isAppleM1 = stdenv.isDarwin && stdenv.isAarch64;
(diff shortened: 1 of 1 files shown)


totten:
Makes sense to me.

And generally -- if you're on a Mac, then you probably have a regular copy of Chrome installed anyway. (And Chrome is relatively stable `#knock-on-wood`.) The managed copy of Chromium is really only needed on CI.

PR #922, opened 2025-04-21
composer.json - Update civicrm-upgrade-test

totten:


totten:
Note: When I run locally, I currently see a failure for upgrading `5.75.2-big` to 6.1. ([Discussion](https://chat.civicrm.org/civicrm/pl/k3wozhcospbf9btsmmq74ocy8h))

PR #927, opened 2025-05-10
Fix find error on macOS

larssandergreen:
Apparently, whether or not `find` works without specifying a path depends on the implementation. On macOS, it does not work and civibuild fails with error `Failed to locate civicrm-core`. This fixes it.


totten:
Oh, yeah. That sounds right - BSD `find` is little less forgiving than GNU `find`. Thanks @larssandergreen.

PR #930, opened 2025-05-25
bknix - When juggling MySQL versions, use separate data folders

totten:
## Before

- MySQL home is `.loco/var/mysql`, which literally stores data

## After

- MySQL home is `.loco/var/mysql/current`, which is a symlink
- MySQL's real folder is `.loco/var/mysql/${MYSQL_ID}` (where the ID is like `mysql57` or `mariadb106`)

## Comment

As written, this would break any existing mysqld folders. Need to think more about upgrade.


totten:
Discussed a bit last week. We'll leave this on ice for a while. Ideally, we could find a way to preserve old builds - e.g. moving data-files into the new locations.

PR #933, opened 2025-05-26
address deprecation warning

eileenmcnaughton:
! [NOTE] Option --cms-base-url is a deprecated alias for --url (-l)


totten:
@eileenmcnaughton Good idea, but looks like a merge-conflict with the contemporaneous #935. See #939.

PR #935, opened 2025-05-26
Find path to inject settings from install output

ejegg:
Rather than using the output of a separate cv ev command. This gets more accurate output for standalone, which was previously injecting the header/footer to add pre.d/post.d settings to the wrong file - civicrm.standalone.php rather than private/civicrm.settings.php.


ufundo:
Thanks for looking at this @ejegg 

I think this could also work and avoids the need for a tmp file:

```
local settings=$( cv ev 'echo CIVICRM_SETTINGS_PATH;' )
if [[ "$settings" == *"civicrm.standalone.php" ]]; then
    settings=$(cv ev 'global $settingsPath; echo $settingsPath;')
fi
```

Ultimately it would be nice if we could get `cv` to set the settings path, well, to the settings path...

PR #937, opened 2025-05-26
vagrant - Drop all support

totten:
This hasn't been updated since PHP 5.x...

I'll need to send a corresponding update to the Dev Guide.


totten:
Related update to devdocs: https://lab.civicrm.org/documentation/docs/dev/-/merge_requests/1222

PR #948, opened 2025-06-06
Convert src/pogo/* to tools/*

totten:
Overview
----------

This shuffles several of buildkit's built-in scripts into a different folder-structure, which will likely appear more familiar to PHP application-developers. 

Before+After
-------------

|Aspect | Before | After
|-- | -- | --
|Script Location | Single file in `src/pogo/XXX.php` | Folder in `tools/XXX/`
|Dependency Declarations| Inline | `tools/XXX/composer.json`
|Dependency Storage | `extern/XXX-php-VER`  | `tools/XXX/vendor`
|Stub References| `bin/XXX` <br/> ⇒ `src/pogo/XXX.php` | `bin/XXX` <br/> ⇒ `tools/XXX/run.php`
|Setup Logic | Bash code in `civi-download-tools` | PHP script (`tools/install.php`) fired by top-level `composer install`

For example:

* __Before__: `bin/fetch-universe` <=> `src/pogo/fetch-universe.php` <=> `extern/fetch-universe-php-8.3/`
* __After__: `bin/fetch-universe` <=> `tools/fetch-universe/` with files `run.php composer.json composer.lock vendor`

Comments
----------

The general motivation is to trim `civi-download-tools` and make the subunits of `builkit.git` more discrete.

Moving toward ^^ required _something_. It doesn't have to be this. (*I had an alternative approach that would continue using pogo -- and I almost went down that path.*) But my guess is that other folks (cc @demeritcowboy @seamuslee001) will be happier with this layout.

Other comments:

* In ancient times, these scripts all used one shared `composer.json` for dependencies. Updates were painful -- someone would bump up package X for tool A, validate tool A, and then break tool B. In their essence, these scripts are independent entities -- IMHO, there's no world where it makes sense for `fetch-universe` and `civicredits` and `zipdiff` to share the same dependency metadata. (But at the same time, they're not quite big enough to justify the overhead of managing independent projects.)
* I've actually been fairly happy with the `pogo` arrangement from a runtime POV and dependency-management POV. (It solved some gnarly dependency-resolution issues around php72 vs php83.) OTOH, it *is* annoying to edit the scripts in an IDE (or to use breakpoints), and it is out-of-mainstream. Some of the scripts would be nicer if split into smaller PHP files. There are silver-linings to a more traditional layout.
* In this moment (following the expulsion of php71+php72), the scripts all get similar dependency-resolutions across the spread (php74...php84).


demeritcowboy:
It seems like an improvement but I'm mostly neutral unless it's going to back to having `universe` as just a single script you can download and run without requiring composer or any code edits. But I don't set up universe that often, or upgrade the script often.

Does src/universe.php still have to be in `[root]/src`? Moving it into the same folder as the script would avoid one edit.

totten:
> Does src/universe.php still have to be in `[root]/src`? Moving it into the same folder as the script would avoid one edit.

Correct, in this case, it can be moved. Pushed that up (and some other cleanups).

totten:
I was hesitating a little because the updated scripts drop support for php73 -- but after grepping some of the job configurations, I don't think it matters.

So... yolo?

PR #958, opened 2025-06-19
Fix Joomla5 zip building

seamuslee001:
This fixes the joomla5 building following https://github.com/civicrm/civicrm-core/commit/a7e5f9668b367385b2fb479c827a57c13935ae8b


totten:
@seamuslee001 deployed

PR #959, opened 2025-06-20
site-list - Convert from plain bullet-list to Bootstrap table

totten:
# Overview

Update the layout of the "Site List" page. This makes it easier to skim, and it also works-around the weird copy-paste-space bug in Firefox.

cc @ufundo @artfulrobot 

# Before

Screenshot 2025-06-20 at 2 43 48 PM

# After

Screenshot 2025-06-20 at 2 48 41 PM


totten:
Or actually...

Screenshot 2025-06-20 at 3 02 01 PM

PR #960, opened 2025-06-24
nixpkgs - Swap v24.05 to v25.05

totten:
The main effects of this are:

1. Bring in new patch revisions of PHP and MySQL.
2. Make better use of the standard CDNs. (These revisions are in the mainline binary-cache.)


totten:
I've been using this a bit today with a few different configurations (new builds, old builds, a few diff test-suites). Seems OK.

totten:
I confirmed that this is causing some test-failures in locale-sensitive tests: https://test.civicrm.org/duderino/batch/CiviCRM-PR/20878237 -- and it reproduces locally.

However, it's a bit strange. The locales do seem to be available:

```bash
$ locale -a | grep ar_AE
ar_AE
ar_AE.iso88596
ar_AE.utf8

$ php -r 'print_r(ResourceBundle::getLocales(""));' | grep ar_AE

    [11] => ar_AE
```

Depending on the environment, the [testLocalizeConstants](https://test.civicrm.org/job/CiviCRM-Test/85553/testReport/(root)/CRM_Utils_DateTest/testLocalizeConstants/) seems a bit chaotic. The CI failure is tiny (`ene.` vs `ene`) and suggests it's still being localized (with different dataset). But on other systems, it's not doing that much.

Feels like... something has qualitatively changed in how the newer `php` package resolves localized data..

totten:
More specifically, it flips between icu69 and icu70.

PR #962, opened 2025-06-25
(infra/ops#1050) CI - Autorelease inactive homerdo mount-points

totten:
Full details: https://lab.civicrm.org/infra/ops/-/issues/1050

Tldr: If you start a job in Jenkins-CI, and if you kill the job before it finishes, then some resources are not properly released. In particular, _processes_ (PIDs) will be killed, but _mount-points_ (e.g. the `tmpfs` where the job does work) are more complicated. They cleanup sometimes - but not always.

This adds a more proactive cleanup procedure. Whenever a server starts a new job, it will first check the statuses for its image files and auto-release any which appear to be unused.

### DIFF
--- a/bin/homerdo
+++ b/bin/homerdo
@@ -169,6 +169,7 @@
     echo "  $prog [OPTIONS] enter [--] [CMD]"
     echo "  $prog [OPTIONS] status"
     echo "  $prog [OPTIONS] kill"
+    echo "  $prog [OPTIONS] auto-release"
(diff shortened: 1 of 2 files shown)


totten:
I've deployed this on `test-4`. Let's have it run for a day on `test-4` before merging/deploying elsewhere.

totten:
No obvious side-effects on `test-4`.

PR #965, opened 2025-07-04
civibuild.aliases.sh - Add aliases for standalone with STABLE/RC/ESR

totten:
Based on suggestions from https://chat.civicrm.org/civicrm/pl/zprzowtdg3fpzehr1h9qbucmcc

I'm not certain if we'll do STABLE/RC/ESR sites for the various CMS's. (It is more sites...) If we do, then we probably want aliases like...

* `d10-stable`
* `d10-rc`
* `wp-stable`
* `wp-rc`

But for the moment, I think we're mostly interested in standalone... and if that's it, then we can have simpler names:

* `stable`
* `rc`
* `esr`

As in:

```bash
civibuild create stable
civibuild create rc
civibuild create esr
```

### DIFF
--- a/src/civibuild.aliases.sh
+++ b/src/civibuild.aliases.sh
@@ -80,6 +80,10 @@ function civibuild_alias_resolve() {
     smaster)     SITE_TYPE=standalone-clean ; CIVI_VERSION=master    ; CMS_TITLE="CiviCRM Standalone Sandbox"        ;;
     sdmaster)    SITE_TYPE=standalone-dev   ; CIVI_VERSION=master    ; CMS_TITLE="CiviCRM Standalone Sandbox"        ;;
 
+    stable)     SITE_TYPE=standalone-clean  ; CIVI_VERSION=stable    ; CMS_TITLE="CiviCRM Standalone Sandbox (Stable)" ;;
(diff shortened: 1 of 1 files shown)


mlutfy:
Can we add "master" for consistency? (master = smaster)

Should builds use `standalone-clean` (composer) or `standalone-dev` (git) 

I was poking around the docs, since I often look up the docs for build types (and this PR will make it easier to memorize), and opened this MR: https://lab.civicrm.org/documentation/docs/dev/-/merge_requests/1229

but that made me wonder about git vs composer for Standalone.

totten:
> Can we add "master" for consistency? (master = smaster)

Good idea. Pushing that up.

> Should builds use `standalone-clean` (composer) or `standalone-dev` (git) 

Erg, docs are outdated there. I believe those were the definitions for a few months in early 2024. Since Jun 2024, the names have been like:

* `standalone-clean` (similar file-layout to the tarball; but based on `git clone` instead of `curl | tar`)
* `standalone-composer` (based on composer for site-building)

IMHO, status quo is: (1) the tarball/`standalone-clean` layout is canon; (2) `composer` site-building is possible for more advanced sysadmins who prefer it.

mlutfy:
Aha ok thanks - I'll re-test and update docs.

PR #968, opened 2025-07-08
give demo user staff role not admin on standalone-* builds

ufundo:
@eileenmcnaughton was keen to have a less permissioned user for testing, e.g. on `smaster`

However - I'm not sure what the password for `admin` user is on `smaster` if you need to do admin things? With this change demo user can't e.g. enable extensions

### DIFF
--- a/app/config/standalone-clean/demo-user.php
+++ b/app/config/standalone-clean/demo-user.php
@@ -35,6 +35,6 @@ CRM_Core_Transaction::create()->run(function () {
     ->addValue('username' , $demoUser)
     ->addValue('password' , $demoPass)
     ->addValue('contact_id' , $contactID)
-    ->addValue('roles:name', ['admin'])
+    ->addValue('roles:name', ['staff'])
     ->execute();
 });
--- a/app/config/standalone-composer/demo-user.php
+++ b/app/config/standalone-composer/demo-user.php
@@ -35,6 +35,6 @@ CRM_Core_Transaction::create()->run(function () {
     ->addValue('username' , $demoUser)
     ->addValue('password' , $demoPass)
     ->addValue('contact_id' , $contactID)
-    ->addValue('roles:name', ['admin'])
+    ->addValue('roles:name', ['staff'])
     ->execute();
 });


eileenmcnaughton:
@ufundo yeah - that's a problem I had too - on the dmaster site we have some text on the landing page to give the 3 user names & passwords

I presume @totten / @seamuslee001 the admin password is not passed in on the command line for smaster but is for dmaster?

totten:
> (@ufundo) However - I'm not sure what the password for admin user is on smaster if you need to do admin things?...
>
> (@eileenmcnaughton) I presume ... the admin password is not passed in on the command line for smaster but is for dmaster?...

These may be a couple questions.

* For the public `smaster.demo.civicrm.org`, the nightly run of https://test.civicrm.org/view/Sites/job/demo.civicrm.org/ will reveal randomly generated password.
* For `*.demo.civicrm.org`, the nightly reset often keeps the same password. (It's not exactly guaranteed on a policy level; but practically, it doesn't change.)
* When creating a new site with `civibuild`, it defaults to a random password - but you can set the admin-password. It works the same way for the Drupal and Standalone builds -- i.e.
    * You can set it explicitly (`--admin-pass s3cr3t`).
    * You can store a default value in your environment (`CIVIBUILD_ADMIN_PASS=s3cr3t`).
    * (There are several other options for `--admin-XXX` and `--demo-XXX` which should also work the same across most ordinary build-types.)

> (@ufundo) eileenmcnaughton was keen to have a less permissioned user for testing, e.g. on smaster
>
> (@eileenmcnaughton) on the dmaster site we have some text on the landing page to give the 3 user names & passwords

I think three users is probably better.

In terms of aesthetics/mnemonics, I don't have a strong feeling about whether `demo` should more sensibly mean "admin" or "staffer". (They're equally reasonable.)

But there is also a functional consideration -- some E2E tests use the `DEMO_USER`, and that's true for all important branches (eg ESR/RC/stable/master). This kind of relies on `DEMO_USER` having similar rights in each build-type. I spot-checked a few tests, and I think this one will have failures if we drop the `demo` user from `role=>admin` to `role=>staffer`:

```
ext/afform/mock/tests/phpunit/E2E/AfformMock/MockPublicFormTest.php
```

Doing a third-user for the `staffer` (like in `drupal-demo`'s advisor Jenny) might be mean less juggling overall?

eileenmcnaughton:
@totten 

"But in practice, we're granting all CiviCRM permissions and ACLs to both users (admin and demo).

    Which maybe means we're granting too much... but conversely, if you're looking for administrative user on Standalone, then demo should do it.. Is there actually any task that demo is blocked from doing on smaster.demo.civicrm.org?
"

I agree that currently demo can do anything - which is what I think we need to change - but I think it's gonna be waay easier on the demo sites if we can use consistent passwords demo/demo admin/admin

I would *definitely* use role based user names - I think the top list is good but I'd probably choose 'volunteer' over 'constituent' - the latter seems to be a term that is used quite a bit in the US but less elsewhere & volunteer fits with 'staff'

Actually - the hierarchy is probably more like
- admin
- staff
- volunteer
- member

But that would only be if we are prepared to let go of our demo/demo muscle memory & also I think demo/demo kinda hints 'try this user'

ufundo:
Will close for now. I think this should go into a new `standalone-demo` build type.

We can an afform at /civicrm/home to display passwords, and coul put some other cool demo stuff there too

PR #971, opened 2025-07-15
Set CMS_VERSION in install.sh in case download.sh didn't set it

highfalutin:
CMS_VERSION is needed by the function civicrm_enable_riverlea_theme. If Drupal has already been downloaded, the variable may not be set.

### DIFF
--- a/app/config/drupal10-demo/install.sh
+++ b/app/config/drupal10-demo/install.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+[ -z "$CMS_VERSION" ] && CMS_VERSION='^10'
+
 ## install.sh -- Create config files and databases; fill the databases
 [ -d "$WEB_ROOT/web" ] && CMS_ROOT="$WEB_ROOT/web"
 


highfalutin:
this may also need to be done in the other configs, but this was the one I was working with.

totten:
> If Drupal has already been downloaded, the variable may not be set.

Interesting. In ordinary workflows, that shouldn't be possible - the value of `CMS_VERSION` is retained via `build/NAME.sh`. If you merely separate the download and install steps, it still carries-forward the `CMS_VERSION` -- the `install` still runs `civicrm_enable_riverlea_theme`.

The following works fine (before/without the patch):

```bash
# Perform download
$ civibuild download tmp --type drupal10-demo

# Check that CMS_VERSION was saved
$ grep CMS_VERSION build/tmp.sh 
CMS_VERSION="^10"

# Perform installation
$ civibuild install tmp
...
Enabling extension "riverlea"
+--------+---------------+-----------+----------+
| scope  | key           | value     | layer    |
+--------+---------------+-----------+----------+
| domain | theme_backend | "minetta" | explicit |
+--------+---------------+-----------+----------+
...
```

I suspect you've stumbled into an edge-case around a prior error. Specifically, if the original download attempt encountered a fatal, then the `build/NAME.sh` may not have been saved -- and so subsequent calls to `civibuild install` don't restore variables (such as `CMS_VERSION` or `ADMIN_PASS` or `DEMO_EMAIL`). I'll open #972 to present that in more detail.

@highfalutin Could it be that you hit the CMS_VERSION error as a downstream side-effect of a prior failure?

PR #979, opened 2025-08-07
Add standalone-demo build to use for demo sites

ufundo:
This adds a `standalone-demo` build type.

The expectation is this would be used for https://smaster.demo.civicrm.org

It adds a standalone demo extension, which:
- adds a notification with login details to the login page
- adds a welcome message on the dashboard

It also enables admin ui / search ui / message admin / chart kit out of the box

CC @seamuslee001 @eileenmcnaughton

### DIFF
--- a/app/config/standalone-demo/download.sh
+++ b/app/config/standalone-demo/download.sh
@@ -16,6 +16,10 @@ pushd "$WEB_ROOT/web"
   git_cache_clone civicrm/civicrm-packages                         -b "$CIVI_VERSION" core/packages
 
   civicrm_l10n_setup private
+
+  # add demo specific extensions
+  git_cache_clone civicrm/standalone_demo                         ext/standalone_demo
+  git_cache_clone civicrm/search_kit_report_starter_pack          ext/search_kit_report_starter_pack
 popd
 
 civibuild_apply_user_extras
--- a/app/config/standalone-demo/install.sh
+++ b/app/config/standalone-demo/install.sh
@@ -38,6 +38,12 @@ civicrm_install_cv
 cv setting:set environment=Development
 cv setting:set debug_enabled=1
 
+# enable demo extension
+cv ext:enable standalone_demo
+
+# enable other extensions relevant to demo
+cv ext:enable civicrm_admin_ui civicrm_search_ui chart_kit search_kit_report_starter_pack
+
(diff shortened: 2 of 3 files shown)


ufundo:
Improved this to 
a) add search kit report starter pack
b) download the extensions in the `download` bit, and use buildkit cleverness for git caching

ufundo:
Have also updated the demo extension itself so that the welcome message is a custom afform rather than a dashlet.

So the login screen looks like:

image


Then on login you see:

image


Could add that message to the login screen in place of the popup if we make the login form a component.

mlutfy:
Would this have the potential of becoming an official demo?

If yes, can I nitpick:

- "Welcome to the CiviCRM Demo" (remove "Standalone" and "Site")
  - While we use "Standalone" officially here and there, it feels unnecessary and potentially confusing to some (for many, this will be their first even encounter with CiviCRM)
  - "Site" is not strictly necessary, and can be confusing to some.
- This mentions that it's running the latest, but ideally we would have a demo running 'stable'? (this would be new - we'd host a real demo of CiviCRM, not a dev sandbox)

maybe that's a wee bit of a tangent, off your initial intent. If so, feel free to ignore.

totten:
> > who might know how to switch the smaster build type 😇
>
> At a glance, I couldn't see what determines the variant used for demo sites (https://test.civicrm.org/view/Sites/job/demo.civicrm.org/configure).

Ah, yeah. Basically, to create `smaster.demo.civicrm.org`, the job runs `civibuild create smaster`.

Names like `smaster` and `dmaster` are modeled as [civibuild aliases](https://docs.civicrm.org/dev/en/latest/tools/civibuild/options/). The relevant file is [civibuild.aliases.sh](https://github.com/civicrm/civicrm-buildkit/blob/master/src/civibuild.aliases.sh). I have some more updates to that file hanging out on my local -- after merging this, I'll push some updated aliases.

Speaking of which - let's merge. The `standalone-demo` build is working for me. 👍

totten:
Related: https://github.com/civicrm/civicrm-buildkit/pull/981/commits/5e1482cf7206962abd6d4d4977139aed69867cd3

totten:
jfyi @ufundo - I pushed up some quick hacks so that the URL, username, and password match the actual values provisioned for the site.

The main commit was https://lab.civicrm.org/extensions/standalone_demo/-/commit/1075409c64e44a06a126590065a61afa9d191df3 (and then there were a few preceding commits to lay groundwork).

PR #980, opened 2025-08-08
remove transitional civicrm_enable_riverlea_theme

ufundo:
This served to enable Riverlea in buildkit builds / on test sites etc.

But it's enabled by default on all new installs since 6.0, so this is redundant.

(It will mean if you make buildkit builds of 5.80/81/82 you wont get Riverlea by default, but I think we can live with that.)


ufundo:
Bump @totten :)

PR #982, opened 2025-08-26
homerdo - Isolate and autoclean /tmp

totten:
This updates the file-sharing rules when running test jobs:

* __Before__: Test jobs use the host's `/tmp` folder. Consequently, it survives across multiple test-runs and can accumulate data.
* __After__: Test jobs use their own `/tmp` (alias for `$HOME/.tmp`). Since each job has its own `$HOME`, this means it also has its own `$HOME/.tmp` (aka `/tmp`).

To make this work, we have to hunt-down any dataflows which rely on (1) sharing data between the inside+outside of container and (2) use `/tmp` for the sharing.

NOTE: In some initial/rough test, this seemed OK. Marking draft because it needs more systemic testing.

### DIFF
--- a/bin/homerdo
+++ b/bin/homerdo
@@ -103,7 +103,6 @@
 ##
 ## - Allow running without any persistent images (`homerdo --temp` starts with empty tmpfs)
 ## - Allow layering multiple images (`homerdo -i base.img -i extra.img --temp`)
-## - Replace /tmp and /var/tmp. (Issue: Sharing tar fifo)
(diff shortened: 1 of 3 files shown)


totten:
Seems OK for running `phpunit-e2e` on an isolated test box. I'm going to deploy to `test-3` on a trial basis. If `test-3` is normal for a couple days, then we can roll out to `test-4` and `test-5`.

totten:
Jobs on `test-3` appear to be producing normal results for the past day.

There were still a few chromium folders in `/tmp`, but I think they're all from yesterday (timestamps before this was posted).

PR #1002, opened 2026-01-17
D7 - Experiment with tag1 releases

totten:
From @demeritcowboy via [Mattermost](https://chat.civicrm.org/civicrm/pl/mx4fs39o4p8g8d36ombwucuauo):

> In the future can also consider using the tag1 releases, since will eventually hit php issues: [https://d7es.tag1.com/node/96](https://d7es.tag1.com/node/96)

They seem to do public releases on a delayed schedule. In theory, those would be adequate for CI-purposes (and better than the `ftp.drupal.org` releases).  I took a quick stab at using this. At the moment, I'm seeing:

* On PHP 8.2, it's OK -- just like the last official 7.103.
* On PHP 8.4, the notes for [7.105](https://gitlab.com/tag1consulting/public/drupal/-/releases) indicate PHP 8.4 support. That's not working for me -- I'm getting issues in `drush site-install`. And if the site is already installed, then I get issues with session-management. (Login doesn't work.) 

Aside @seamuslee001  - I didn't realize there was a [drush 8.5](https://github.com/drush-ops/drush/releases/tag/8.5.0). It doesn't help with php84... but do you know if we still need the custom build of `drush.phar`?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
       },
       {
         "title": "Setup local PHP tools",
-        "run": "@php-script tools/install.php bknix-ci-cleanup civicredits civici fetch-universe find-stale-builds forkify import-rn phpunit-xml-cleanup releaser zipdiff"
+        "run": "@php-script tools/install.php bknix-ci-cleanup civicredits civici fetch-universe find-stale-builds forkify gitlab-release import-rn phpunit-xml-cleanup releaser zipdiff"
       }
     ],
     "compile-passthru": "always",
--- a/phars.json
+++ b/phars.json
@@ -41,8 +41,8 @@
    },
 
    "drush8":  {
-     "url": "https://download.civicrm.org/cv/drush.phar-8.4.12",
-     "sha256": "cc19dd738bfb8a6324bc89de5f801a24a758f20c659d66990b881324d29f8922",
+     "url": "https://github.com/drush-ops/drush/releases/download/8.5.0/drush.phar",
+     "sha256": "d814ec11c9990f2f95fc6e91be77ec6f5deb71d1980142c3f8ce3ad651fd1bae",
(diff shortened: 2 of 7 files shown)


demeritcowboy:
Regarding their release schedule, yes it's delayed from the paid version but I can tell you that at this moment there is nothing in the paid version that isn't in the public version for drupal core, so this wouldn't be missing any known patches. For contrib modules there might be but that shouldn't matter for initial testing here.

Regarding 8.4, I don't have any insight to add right now since I haven't really given it a good try for anything.

demeritcowboy:
I did get an install working with php 8.4, both with drush 8.5 and the UI. But I had to hack drupal core, and then leave my hack in after still to avoid other crashes. And then there's numerous warnings about "implicit null" function params, which in turn with drush lead to warnings about "Session ini settings cannot be changed after headers have already been sent", which I assume then cause any `ini_set(session....` to get ignored (that didn't cause me any problems but I didn't test out drush much with other commands).

For the record:
* includes/entity.inc
  * line 23: `?array`
  * line 143: `?array`
  * why the above 2 only and not the many others in the code? I dunno. It does an include instead of require on this file, and maybe it's more sensitive to the warning output, but even with error_reporting=1 it still didn't work without it.
* includes/bootstrap.inc
  * line 2691 in `_drupal_exception_handler` change `_drupal_log_error(_drupal_decode_exception($exception), TRUE);` to something that doesn't output and gives more info, like `error_log($exception->getMessage() . ' ' . $exception->getTraceAsString());`. This is how I found the entity.inc issue because `_drupal_decode_exception` didn't point to it.
* includes/errors.inc
  * line 67 in `_drupal_error_handler_real` replace `_drupal_log_error()` with e.g. `error_log($message . ' ' . print_r($caller, true));`.

PR #1061, opened 2026-06-19
phars.json - Update drush8 to v8.5.0

totten:
This de-forks the copy of drush8. The old/custom build reports its `--version` as `8.4.12-dev`. Current upstream is 8.5.0.

@seamuslee001 Do you remember what the issue was that required the custom build?

Anecdotally... I'm able to  local runs of `civibuild reinstall dmaster` and `civibuild reinstall bcmaster`, and this fixes a problem in running `civibuild install bluebird`

### DIFF
--- a/phars.json
+++ b/phars.json
@@ -41,8 +41,8 @@
    },
 
    "drush8":  {
-     "url": "https://download.civicrm.org/cv/drush.phar-8.4.12",
-     "sha256": "cc19dd738bfb8a6324bc89de5f801a24a758f20c659d66990b881324d29f8922",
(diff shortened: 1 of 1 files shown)


seamuslee001:
@totten I can only assume it was some PHP version compatability issue I think but not 100% sure at all

totten:
Whatever it was, I'm pretty sure you would've sent it upstream, and poking through https://github.com/drush-ops/drush/pulls?q=is%3Apr+author%3Aseamuslee001+is%3Aclosed shows they're all resolved now.

Also, did a `civibuild reinstall d10`, and that seemed OK.

PR #1091, opened 2026-08-22
Universe - Exclude archived repos

colemanw:
Removes clutter in `fetch-universe`.

Note: https://github.com/civicrm/civicrm-buildkit/pull/995 was supposed to exclude "unsupported" projects but that was effectively undermined by the gitlab feed which doesn't care. This at least gets the archived projects out of the feed.

But still we're left with only excluding projects that are **Unsupported AND Archived**.
My preference would be to exclude **Unsupported OR Archived**, but at least this PR is a step in the right direction.

Prior to this PR it would only have excluded projects that are **Unsupported AND not hosted on gitlab** so this is definitely better.


demeritcowboy:
It seems only the core team are owners of projects in labs, and so even maintainers of extensions can't archive them. We can manage the topics and the url filter supports "topic=" but I don't see how to do negative url filters, but topics is in the response so the list could be post-filtered in the code here.

demeritcowboy:
Or the procedure would be we could ask a core team member to do it. How often does it come up.

colemanw:
> Or the procedure would be we could ask a core team member to do it. How often does it come up.

Yea, for now. At some point we might want to revisit that but at the moment it comes up very rarely.

